PR #3, opened 2020-09-08
Add React and dependencies

Erin007:
Screen Shot 2020-09-08 at 1 32 00 PM


package.json
```diff
@@ -14,16 +14,21 @@
   },
   "dependencies": {
     "babel-core": "^6.26.3",
-    "babel-loader": "^8.1.0",
     "babel-polyfill": "^6.26.0",
     "babel-preset-env": "^1.7.0",
     "babelify": "^10.0.0",
     "eslint-plugin-react": "^7.20.3",
-    "webpack": "^4.43.0",
+    "react": "^16.13.1",
```
breville, line 21: I suspect we might want to use the same version of React as https://studio.code.org does, particularly for when we are running as hosted content.  I notice that our Oceans repo does [that](https://github.com/code-dot-org/ml-activities/blob/main/package.json#L38).

package-lock.json
```diff
@@ -13,36 +13,1298 @@
         "@babel/highlight": "^7.10.4"
       }
     },
+    "@babel/compat-data": {
```
breville, line 16: I'm not sure that we want this file in our repo?

breville (approved):

PR #4, opened 2020-09-14
import data from CSV

Erin007:
Using [papaparse](https://www.papaparse.com/docs), we can now upload a csv file and store the data from the file as state. 
 
ml-import-csv


src/CSVReaderWrapper.jsx
```diff
@@ -0,0 +1,64 @@
+import React, { Component } from "react";
+import Papa from "papaparse";
+
+export default class CSVReaderWrapper extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
breville, line 8: I guess we can hoist this state up to a common state object soon...

src/CSVReaderWrapper.jsx
```diff
@@ -0,0 +1,64 @@
...
+      csvfile: undefined,
+      data: undefined
+    };
+
+    this.updateData = this.updateData.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+    this.importCSV = this.importCSV.bind(this);
```
breville, line 15: I think ES6 lets you use `=>` instead of `.bind(this)`...

breville (approved):
Exciting first steps!

PR #5, opened 2020-09-17
Implement SVM ML model to analyze example candy data set 

Erin007:


src/CSVReaderWrapper.jsx
```diff
@@ -25,10 +39,85 @@ export default class CSVReaderWrapper extends Component {
     });
   };
 
+  extractLabels = row => {
+    return parseInt(row[this.state.labelColumn]);
+  };
+
+  extractExamples = row => {
+    let exampleValues = [];
+    this.state.selectedFeatures.forEach(feature =>
```
breville, line 48: Could we use `.map` here?

breville:
As dicsussed, one architecture we might want to move towards is decoupling the third-party `svm` library from our UI code.  If we were able to hoist it into its own model, then it would be easier to unit test our SVM functionality (independent of which third-party library we use), and also easier to use a different or additional ML library with minimal impact on our UI code.

PR #7, opened 2020-09-24
Test: setup first test

breville:
I adjusted the Node version since https://nodejs.org/en/about/releases/ lists 14 as the current version right now.


Erin007 (approved):
Yay! Thanks for setting this up. Node version bump seems fine.

PR #8, opened 2020-09-30
Have user indicate what kind of data - continuous or categorical - is in each column

Erin007:
[STAR-1285](https://codedotorg.atlassian.net/browse/STAR-1285)

The bulk of the work here is to allow the user to indicate what kind of data - categorical, continuous or other - is stored in each of the columns of their imported dataset. Having the user provide this info 1.) sets us up to do the conversion of string categories to numerical categories for categorical data more easily 2.) sets us up to exclude columns that don't make sense for labels and features to train on 3.) eventually, potentially sets us up to recommend algorithms based on the type of data in the dataset.

Screen Shot 2020-09-29 at 8 42 51 PM

I accomplished this by setting up a new piece of state `metaDataByColumn` which includes a field for dataType. Next I'll expand that piece of state to include information about unique values stored in each categorical column and their numerical keys. 

Also included is [Star-1270](https://codedotorg.atlassian.net/browse/STAR-1270) - placeholder explanations of label and feature to increase usability of the prototype .

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -61,9 +61,6 @@ export default function rootReducer(state = initialState, action) {
     };
   }
   if (action.type === SET_METADATA_BY_COLUMN) {
-    const prevMetaData = state.metaDataByColumn[action.column]
-      ? state.metaDataByColumn[action.column][action.metadataField]
-      : {};
     return {
       ...state,
       metaDataByColumn: {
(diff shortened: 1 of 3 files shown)


src/redux.js
```diff
@@ -51,6 +60,21 @@ export default function rootReducer(state = initialState, action) {
       data: action.data
     };
   }
+  if (action.type === SET_METADATA_BY_COLUMN) {
+    const prevMetaData = state.metaDataByColumn[action.column]
```
breville, line 64: What is `prevMetaData` intended to be used for?

Erin007: nothing! It was leftover from a different path I tried. Deleted! Thanks for catching.

breville (approved):

PR #9, opened 2020-09-30
Constrain which columns can be selected based on data type 

Erin007:
[STAR-1264](https://codedotorg.atlassian.net/browse/STAR-1264)
[STAR-1265](https://codedotorg.atlassian.net/browse/STAR-1265)

Now that have metadata about which columns contain which data types - continuous, categorical or other. I used that information to constrain selections for training to guide users to more reliably being able to train the model. The constraints I implemented are: 
- Only continuous and categorical columns can be selected as feature columns.
- Only categorical columns can be selected as the label column. 
- If a column is selected as the label column, it can't also be selected as a feature column.  

constrain-dropdowns-by-datatype

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -3,7 +3,7 @@ import _ from "lodash";
 // Action types
 
 const SET_IMPORTED_DATA = "SET_IMPORTED_DATA";
-const SET_METADATA_BY_COLUMN = "SET_METADATA_BY_COLUMN";
+const SET_COLUMNS_BY_DATA_TYPE = "SET_COLUMNS_BY_DATA_TYPE";
 const SET_SELECTED_FEATURES = "SET_SELECTED_FEATURES";
 const SET_LABEL_COLUMN = "SET_LABEL_COLUMN";
 const SET_SHOW_PREDICT = "SET_SHOW_PREDICT";
@@ -16,11 +16,10 @@ export function setImportedData(data) {
   return { type: SET_IMPORTED_DATA, data };
 }
 
-export const setMetaDataByColumn = (column, metadataField, value) => ({
-  type: SET_METADATA_BY_COLUMN,
+export const setColumnsByDataType = (column, dataType) => ({
+  type: SET_COLUMNS_BY_DATA_TYPE,
   column,
-  metadataField,
-  value
+  dataType
 });
 
 export function setSelectedFeatures(selectedFeatures) {
@@ -45,7 +44,7 @@ export function setPrediction(prediction) {
 
 const initialState = {
   data: [],
-  metaDataByColumn: {},
+  columnsByDataType: {},
   selectedFeatures: [],
   labelColumn: undefined,
   showPredict: false,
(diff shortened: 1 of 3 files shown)


src/redux.js
```diff
@@ -109,6 +111,35 @@ export function getFeatures(state) {
   return state.data.length > 0 ? Object.keys(state.data[0]) : [];
 }
 
+function filterColumnsByType(state, columnType) {
+  let columnsOfType = [];
+  if (!_.isEmpty(state.metaDataByColumn)) {
+    var filteredObject = _.pickBy(state.metaDataByColumn, function(
```
breville, line 117: nit: can possibly use arrow operator instead of `function` here?

src/redux.js
```diff
@@ -109,6 +111,35 @@ export function getFeatures(state) {
...
+export function getCategoricalColumns(state) {
+  return filterColumnsByType(state, ColumnTypes.CATEGORICAL);
+}
+
+export function getContinuousColumns(state) {
+  return filterColumnsByType(state, ColumnTypes.CONTINUOUS);
+}
+
+export function getSelectableFeatures(state) {
+  return _.pull(
```
breville, line 137: nit: this mutates the array in place, and while possibly more efficient, it might be good to use immutable alternatives whenever possible?

breville:
Just curious: from playing with the UI, it's neat that something disappears from a list if its data type is changed to something that removes it from that list, but changing the data type back shows it selected in that list again.  Does that mean the state isn't fully cleared when a data type is changed?

src/redux.js
breville: It sounds like you're already doing this for most of the data flow, but just to capture the idea: given all the different ways we manipulate the original CSV-based data, it would be good to treat that data as immutable and to always have a chain of transformations that generate new views of that original data, rather than ever modifying the original data.

src/redux.js
```diff
@@ -1,3 +1,5 @@
+import _ from "lodash";
```
breville, line 1: As discussed, it might be good to decide whether we still need to use lodash given all the latest JS functions.  It's fine to continue using it, but it seems worth making a conscious decision to only use it when it's better than internal JS functionality.

breville:
As discussed, it might be good to find a way to avoid having to re-setup configuration each time a CSV is imported.  One solution, at least for internal use, might be to allow annotation on column names to declare data types, etc.

Erin007:
> Just curious: from playing with the UI, it's neat that something disappears from a list if its data type is changed to something that removes it from that list, but changing the data type back shows it selected in that list again. Does that mean the state isn't fully cleared when a data type is changed?

I updated based on the following "rules": 
- if a column is selected as the label column and its data type changes to "other" or "continuous" it's un-selected 
- if a column is selected as a feature to train on and its data type changes to "other" it's un-selected

Erin007:
> As discussed, it might be good to find a way to avoid having to re-setup configuration each time a CSV is imported. One solution, at least for internal use, might be to allow annotation on column names to declare data types, etc.

That sounds handy.  We could modify the `setDefaultColumnDataType` function in `CSVReaderWrapper` to handle this; but I'll save for a separate PR.

src/redux.js
Erin007: I removed all the references to lodash. Filter does the trick and doesn't mutate the og array.

breville (approved):

PR #11, opened 2020-10-05
Strings to numbers for svm

Erin007:
This sets it up so that categorical string data can be converted into numbers for training and testing using the current binary classification SVM model. 

_Some notes on the terminology I used:_ 

Column headers are the names of features, thus "**column**" and "**feature**" are interchangeable. For example, in the Titanic dataset, "survived", "sex", "pclass" etc are features aka columns. 

**options** refers to the categories within a column. For example, the "options" for the "pclass" feature column in the Titanic dataset has options: 1, 2, and 3 to indicate which class the passenger traveled in. 

An example of `state.featureNumberKey` for a string-based categorical dataset might look like:
 
```
 state.featureNumberKey = {
   sex: { male: 0, female: 1 },
  favoriteColor: {green: 0, blue: 1, yellow: 2, red: 3},
  birthMonth: {january: 0, february: 1, march: 2 ... } 
}
```

### DIFF
--- a/src/SVMTrainer.js
+++ b/src/SVMTrainer.js
@@ -3,7 +3,12 @@
 /* eslint-env node */
 const svmjs = require("svm");
 import { store } from "./index.js";
-import { setPrediction } from "./redux";
+import {
+  setPrediction,
+  getUniqueOptions,
+  getSelectedCategoricalColumns,
+  setFeatureNumberKey
+} from "./redux";
 
 export default class SVMTrainer {
   constructor() {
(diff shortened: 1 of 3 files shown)


src/SVMTrainer.js
```diff
@@ -14,20 +20,59 @@ export default class SVMTrainer {
...
     const state = store.getState();
     this.state = state;
+    this.buildFeatureToNumberKey(state);
   }
 
   startTraining() {
     this.addTrainingData();
   }
 
+  buildFeatureToNumberKeyForColumn = column => {
```
breville, line 30: Could we get a small comment for each function?  And maybe one bigger comment somewhere documenting how this new feature ties together?

src/SVMTrainer.js
```diff
@@ -14,20 +20,59 @@ export default class SVMTrainer {
...
+    let optionsMappedToNumbers = {};
+    const uniqueOptions = getUniqueOptions(this.state, column);
+    uniqueOptions.forEach(
```
breville, line 33: Any chance `.map` would work for these?

src/SVMTrainer.js
Erin007: I added comments and fleshed out the description in the PR. Anything missing that would help?

Erin007:
Also, just for posterity... I was missing the `reduxLogging` setting we have in the main repo, but I learned that you can look at nested objects in the console with: 

``` 
const util = require("util");
console.log(util.inspect(<nestedObect>, { depth: # }))
```

src/SVMTrainer.js
```diff
@@ -14,24 +20,118 @@ export default class SVMTrainer {
...
+    categoricalColumnsToConvert.forEach(
+      feature =>
+        (optionsMappedToNumbersByFeature[feature] = this.buildOptionNumberKey(
+          feature
+        ))
+    );
+    store.dispatch(setFeatureNumberKey(optionsMappedToNumbersByFeature));
+  }
+
+  /* Builds a hash that maps the options in the label column to -1 and 1, because right now the model only supports binary classification with those specific values. Returns the row's specific option's value in that reference hash.
```
breville, line 73: nit: very long line of text.

breville (approved):
Thanks for documenting.  This is great!

PR #13, opened 2020-10-08
Dataset manifest to pull in remote data from CSVs in the repo 

Erin007:
Follow up to #12 
dataset-collection-selection


breville (approved):
With caveats that we should work out why the csv files are showing even though they were in a previous PR, the UI could maybe clarify that the upload button needs to be pressed after choosing from the dropdown (or maybe we could upload as soon as it's chosen?), and we should be aware that this will always pull the published file rather than the latest local copy, then this seems good :)

PR #16, opened 2020-10-17
Three new datasets: food, flight delays, videogame sales

breville:
These are served by the active webserver.


Erin007:
Do you think I should stop referencing github for the other datasets?

Erin007 (approved):

breville:
Very likely we should, since it means each environment will pull its local, latest copy of the data.  But we can see if all environments work properly in this PR before switching the rest.

PR #23, opened 2020-10-23
Add basic pre-training validation

Erin007:
[AI-25](https://codedotorg.atlassian.net/browse/AI-25)

validation

### DIFF
--- a/src/UIComponents/TrainModel.jsx
+++ b/src/UIComponents/TrainModel.jsx
@@ -86,22 +86,34 @@ class TrainModel extends Component {
                   the newly trained model.
                 </p>
                 <div>
-                  <table>
-                    <tr>
-                      <th>Expected</th>
-                      <th>Predicted</th>
-                    </tr>
-                    {this.props.accuracyCheckLabels.map((label, index) => {
-                      return (
+                  <p>
+                    10% of the training data was reserved to test the accuracy
+                    of the newly trained model.
+                  </p>
+                  <div>
+                    <table>
+                      <thead>
                         <tr>
-                          <td>{label}</td>
-                          <td>
-                            {this.props.accuracyCheckPredictedLabels[index]}
-                          </td>
+                          <th>Expected</th>
+                          <th>Predicted</th>
                         </tr>
-                      );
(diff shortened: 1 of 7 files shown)


src/UIComponents/TrainModel.jsx
```diff
@@ -34,53 +50,94 @@ class TrainModel extends Component {
...
-                going to look for patterns in these features:{" "}
-                {this.props.selectedFeatures.join(", ")} that might help predict
-                the values of the label: {this.props.labelColumn}.
-              </p>
-              <button type="button" onClick={this.onClickTrainModel}>
-                Train model
-              </button>
-              {this.state.showAccuracy && (
+        <h2>Are you ready to train the model?</h2>
+        {readyToTrainErrors.noFeature && (
```
breville, line 57: I wonder if we could create a JS data structure which maps these errors to their error strings?

src/redux.js
```diff
@@ -354,6 +354,32 @@ export function getAccuracy(state) {
   return ((numCorrect / numPredictedLabels) * 100).toFixed(2);
 }
 
+export function readyToTrain(state) {
+  return !Object.values(readyToTrainErrors(state)).includes(true);
+}
+
+export function readyToTrainErrors(state) {
+  const readyToTrainErrors = {};
+  readyToTrainErrors["noFeature"] = state.selectedFeatures.length === 0;
```
breville, line 363: Can we also just say `readyToTrainErrors.noFeature = `?

breville (approved):

src/UIComponents/TrainModel.jsx
Erin007: oooh that's better! I was wondering about something like that I when I realized in the UI I had to know the specific keys to use to access the info in `readyToTrainErrors`.  I'll update.

breville: Longer term, this kind of conditional feedback reminds me of the "guide" we built for Oceans at https://github.com/code-dot-org/ml-activities/blob/main/src/oceans/models/guide.js.

PR #25, opened 2020-10-28
Publish app

breville:
This sets up the AI Lab app to be published as an npm module and then embedded in https://studio.code.org.  

See https://github.com/code-dot-org/code-dot-org/pull/37495 for the initial usage.

This is mostly based on what was done for Oceans in https://github.com/code-dot-org/ml-activities, which was in turn based on what was done for Dance Party in https://github.com/code-dot-org/dance-party.


public/index.html
```diff
@@ -4,11 +4,10 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <script src="https://www.lactame.com/lib/ml/4.0.0/ml.min.js"></script>
```
Erin007, line 7: I think I need this script to make the KNN algorithm work. Can we add it back or did it cause problems?

Erin007: oh, I see, you moved it into `KNNTrainer` 👍

webpack.config.js
```diff
@@ -1,34 +1,110 @@
...
+    ...externalConfig
+  }
+];
+
+module.exports = (env, argv) => {
+  if (argv.mode === 'production') {
+    return [...defaultConfig, ...productionConfig];
+  }
+
+  console.log("!!" ,defaultConfig);
```
Erin007, line 106: still needed?

webpack.config.js
```diff
@@ -1,34 +1,110 @@
...
+const defaultConfig = [
+  {
+    entry: {
+      assetPath: './src/assetPath.js'
+    },
+    ...commonConfig,
+    ...firstConfigOnly,
+    ...externalConfig
   },
-  plugins: [new webpack.HotModuleReplacementPlugin()]
```
Erin007, line 33: Will it still hotload?

public/index.html
breville: Yes we now use a published version.

src/assetPath.js
```diff
@@ -0,0 +1,5 @@
+global.__ml_playground_asset_public_path__ = './assets/';
+
+export const setAssetPath = (path) => {
+  global.__ml_playground_asset_public_path__  = path;
+};
```
Erin007, line 5: Is this asset path to access the the datasets? I'm not clear on how this part works.

webpack.config.js
breville: I'm not sure.  Was it working before?

Erin007 (approved):
Couple of questions, mostly for my own education, but nothing blocking. Let's see what happens!

src/assetPath.js
breville: Yes, this variable is used to override the asset path by the main repo's code [here](https://github.com/code-dot-org/code-dot-org/pull/37495/files#diff-f2ae9cbcf69d10020839a1ef20c80985272ddde9ed3baf55d36bf532229fe8b6R122). 

If we want to publish assets to be imported by the app we can also use it like this: https://github.com/code-dot-org/ml-activities/blob/main/webpack.config.js#L49.

One subtle caveat is that the variable is reset [here](https://github.com/code-dot-org/ml-playground/pull/25/files#diff-64698619a0c48c473fa8e31b2ebe38f6ac73bbeb7f8a491dd65aed5d67b79579R1) if we `require` this module, say at app load time, so we need to be careful about its usage.

PR #26, opened 2020-10-29
Initial basic data validation

Erin007:
Jira: AI-57

Adds 3 basic pre-training checks to ensure the data is formatted correctly: 

- there must be data (follow up work will check dataset size relative to complexity) 
- columns must be named with unique names
- there shouldn't be any empty cells 

When the data meets all the requirements: 
Screen Shot 2020-10-28 at 8 15 20 PM

When the data does not meet the requirements: 
Screen Shot 2020-10-28 at 8 15 31 PM

I also discovered that there were empty rows appended for some datasets, so I updated the config we're sending to Papa Parse to fix that issue.
Screen Shot 2020-10-28 at 8 07 27 PM

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -2,6 +2,10 @@
 
 import { availableTrainers } from "./train.js";
 
+export function datasetSize(state) {
+  return !!state.data.length > 0;
+}
+
 export function minOneFeatureSelected(state) {
   return state.selectedFeatures.length !== 0;
 }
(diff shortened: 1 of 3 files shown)


src/validate.js
```diff
@@ -2,6 +2,49 @@
...
+export function columnsNamed(state) {
+  const columnNames = getColumnNames(state);
+  return (
+    columnNames.length > 0 &&
+    !columnNames.includes("") &&
+    !columnNames.includes(undefined)
+  );
+}
+
+export function uniqueColumnNames(state) {
```
breville, line 23: Could we have a brief comment describing what each function does?

PR #28, opened 2020-10-29
Show column stats

Erin007:
Jira: AI-38, AI-57

Students are going to need to be able to "inspect" their data to check for errors in preparation for training. 

For categorical columns, we now show the number of unique values in the column and what those values are. For continuous columns, we now validate that the column only contains numerical values and show the max and min values in the column.

Screen Shot 2020-10-29 at 1 40 06 PM

This will help students identify when they have more categories than anticipated, for example: 
Screen Shot 2020-10-29 at 1 17 11 PM

If they've mis-identified the datatype stored in a column or if this there is a string lurking in a column that should only contain numbers, for example: 
Screen Shot 2020-10-29 at 2 40 31 PM

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -2,6 +2,7 @@
 
 import { availableTrainers } from "./train.js";
 import { ColumnTypes } from "./constants.js";
+import { getSelectedContinuousColumns } from "./redux.js";
 
 export function minOneFeatureSelected(state) {
   return state.selectedFeatures.length !== 0;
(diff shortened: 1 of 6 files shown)


src/validate.js
```diff
@@ -36,6 +37,22 @@ export function selectedColumnsHaveDatatype(state) {
   return selectedColumns.length > 0 && columnTypesOk;
 }
 
+// Check that selected continuous columns only contain numbers.
+// @return {boolean}
+export function continuousColumnsHaveOnlyNumbers(state) {
+  const columns = getSelectedContinuousColumns(state);
+  let allNumbers = true;
+  state.data.forEach(function(row, i) {
```
breville, line 45: Is `.forEach` a problem in IE11?

breville (approved):
These functions look like good candidates for unit testing too.

Erin007:
> These functions look like good candidates for unit testing too.

definitely - I opened a [task](https://codedotorg.atlassian.net/browse/AI-55) for this.

src/validate.js
Erin007: for `NodeLists`, but I think it's just fine for Arrays.  I'll research a bit to be sure.

Erin007: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Browser_compatibility

PR #30, opened 2020-11-04
Extract CSV parsing from UI 

Erin007:
In preparation for parameterizing the tools for level building, I extracted the CSV parsing step from the UI and did a little refactoring along the way: 
- The UI component was renamed from `CSVReaderWrapper` to `SelectDataset`
- a new file, `csvReaderWrapper` now contains the helper functions to parse the CSV and set the data in the Redux store, along with default column types 
- `csvfile` is now a piece of state in the Redux store so it can be set both from the `SelectDataset` UI component and from elsewhere via upcoming parameterization
- Renamed `availableDatasets` to `allDatasets` since we'll likely need to take a subset of that list in future parameterization tasks; for example, if a level builder want to expose a limited number of datasets.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import CSVReaderWrapper from "./UIComponents/CSVReaderWrapper";
+import SelectDataset from "./UIComponents/SelectDataset";
 import DataDisplay from "./UIComponents/DataDisplay";
 import SelectFeatures from "./UIComponents/SelectFeatures";
 import ColumnInspector from "./UIComponents/ColumnInspector";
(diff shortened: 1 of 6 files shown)


Erin007:
@breville With this set up you should now be able to have `csvfile` as a parameter. Then from index.js (maybe, or wherever your params are coming in) you could do something like: 

```
store.dispatch(setSelectedCSV(<csvfileParameter>) 
parseCSV(<csvfileParameter>, false)
``` 
Is that what you were thinking? 

Up next, I can hide/show the `SelectData` component conditionally so that will also parameterizable, and then find a way to parameterize  which subset of `allDatasets` shows in the dropdown when `SelectData` is visible.

breville (approved):

PR #32, opened 2020-11-05
Initial work on parameterisation

breville:
This is an initial attempt at simple parameterisation of the app.

`initAll` now takes an optional parameter, and the first one that actually does something looks like this:
```
initAll({ mode: { id: "load_dataset", setId: 3 } });
```
In particular, this loads the "Foods" dataset.  It also hides the dataset selection UI entirely.

For local runs, a dropdown is now shown:

Screen Shot 2020-11-05 at 4 41 24 PM

If an option is chosen, the page is reloaded with a `?mode=` URL parameter for the appropriate "sample mode".

This makes use of the factoring done in https://github.com/code-dot-org/ml-playground/pull/30.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,16 +2,33 @@ import React from "react";
 import ReactDOM from "react-dom";
 import App from "./App.js";
 import { Provider } from "react-redux";
-import { createStore } from "redux";
-import rootReducer from "./redux";
+import { createStore} from "redux";
+import rootReducer, { setMode, setSelectedCSV } from "./redux";
+import { allDatasets } from "./datasetManifest";
+import { parseCSV } from "./csvReaderWrapper";
 
 export const store = createStore(rootReducer);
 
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -2,16 +2,33 @@ import React from "react";
...
+import rootReducer, { setMode, setSelectedCSV } from "./redux";
+import { allDatasets } from "./datasetManifest";
+import { parseCSV } from "./csvReaderWrapper";
 
 export const store = createStore(rootReducer);
 
 export const initAll = function(options) {
+  const mode = options && options.mode;
+
+  setMode(mode);
```
Erin007, line 15: Do you need to do `store.dispatch(setMode(mode));` or is this enough?

breville: Will fix!

Erin007 (approved):
Progress! Wahoo!

PR #40, opened 2020-11-12
Dataset metadata support

breville:
Loading a dataset via parameter now loads its metadata too.  The metadata is defined in a JSON file that sits alongside the data's CSV file.  The metadata controls the feature types, with the UI only showing them, but not allowing them to be selected.

Note also that parameterisation has changed slightly to use a more friendly id for the dataset, e.g.
```
initAll({ mode: { id: "load_dataset", setId: "foods" } });
```

### DIFF
--- a/public/datasets/erin_candy_preferences.json
+++ b/public/datasets/erin_candy_preferences.json
@@ -1,0 +1,59 @@
+{
+  "fields": [
+  	{
+  		"id": "name",
+  		"type": "categorical",
+  		"name": "Name"
+  	},
+  	{
+  		"id": "chocolate",
+  		"type": "categorical",
+  		"name": "Name"
+  	},
+  	{
+  		"id": "fruity",
+  		"type": "categorical",
+  		"name": "Name"
+  	},
+  	{
+  		"id": "caramel",
+  		"type": "categorical",
+  		"name": "Name"
+  	},
+  	{
+  		"id": "nuts",
+  		"type": "categorical",
+  		"name": "Name"
+  	},
+  	{
+  		"id": "nougat",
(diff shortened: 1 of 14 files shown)


public/datasets/erin_candy_preferences.json
```diff
@@ -0,0 +1,59 @@
+{
+  "fields": [
+  	{
+  		"id": "name",
+  		"type": "categorical",
+  		"name": "Name"
```
breville, line 6: I need to go through and set proper types for each of these.

I added a friendly name field, but haven't set it yet.  Do we think it will be useful?

Erin007: I think the friendly name field will be useful, especially for more complicated datasets or when the column name isn't very descriptive. However, I don't think we should worry about setting the friendly name for all the columns for each of these datasets since we likely won't end up using this data.

Erin007: Also, might be interesting to loop Brent in on the general idea of how this metadata will work (once you're feeling like it's stable-ish and ready to share) since I suspect he'll be working on writing up these metadata files for the final sets?

src/redux.js
```diff
@@ -508,7 +532,7 @@ export function validationMessages(state) {
   validationMessages.push({
     readyToTrain: compatibleLabelAndTrainer(state),
     errorString:
-      "The label datatype must be compatiable with the training algorithm.",
+      "The label datatype must be compatible with the training algorithm.",
```
Erin007, line 535: thank you 🙂

Erin007 (approved):

PR #43, opened 2020-11-17
Prepare to save trained models and details

Erin007:
Adds a model name and description field
Screen Shot 2020-11-17 at 10 32 27 AM

Then uses the inputed information in addition with other model details already in state to prepare a bundle of data to pass to the `saveTrainedModel` function. 

Compatible with [Code Studio PR #37811](https://github.com/code-dot-org/code-dot-org/pull/37811)

### DIFF
--- a/src/UIComponents/SaveModel.jsx
+++ b/src/UIComponents/SaveModel.jsx
@@ -20,9 +20,10 @@ class SaveModel extends Component {
   };
 
   onClickSave = () => {
-    console.log("dataToSave", this.props.dataToSave);
     this.props.saveTrainedModel(this.props.dataToSave);
-    alert("Don't get too excited, this doesn't do anything yet :)");
+    alert(
+      "Ok, you can get a little excited now; this saves the trained model to an s3 bucket!"
+    );
   };
 
   render() {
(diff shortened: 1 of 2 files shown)


src/UIComponents/SaveModel.jsx
```diff
@@ -2,23 +2,48 @@
...
   render() {
     return (
       <div>
         <br />
         <br />
+        <label>
+          Model name:
+          <input
+            type="text"
+            onChange={event => this.handleChange(event, "name")}
```
breville, line 38: It might be worth making this a controlled component where the value is set via the state, as described at https://reactjs.org/docs/forms.html

Erin007: Why? Then we'd have to keep the component's internal state consistent with the Redux state coming in as a prop. Am I missing something?

breville: It is likely fine to leave this as-is for now.  

The challenge with uncontrolled components is that they don't follow the React model of having state flow in one direction, and instead there is some state inside the HTML component that is outside of the React/Redux state, and can in fact drift out of sync.  I've had issues with more complex third-party components that have their own internal state which we can't control.

breville (approved):

src/UIComponents/SaveModel.jsx
Erin007: Ah, I see. I think I'll merge as is for the sake of getting something working to play with and note that we can update later.

breville: For what it's worth, https://github.com/code-dot-org/ml-playground/pull/88 is a good example of why controlled is useful.

PR #44, opened 2020-11-25
add cleaned sat scores dataset

brondle:
added a cleaned dataset with a JSON file for categorical/continuous


breville (approved):
Would it be worth committing a "pretty print" JSON or better to leave it like this?

brondle:
> Would it be worth committing a "pretty print" JSON or better to leave it like this?

Oops, didn't even think about pretty printing. I should be adding a few more datasets in on Friday so I can update my pipeline to look better then.

PR #45, opened 2020-12-03
One panel at a time

breville:
This updates the AI Lab UI to show one panel at a time, with a set of navigation tabs across the top, and a checklist of per-panel requirements on the right.

Panels are enabled as their precursor requirements are met; if a panel is disabled, all panels to its right are also disabled.  This gives the the experience of "unlocking" access to panels from left-to-right.

Screen Shot 2020-12-04 at 5 23 17 PM

This change also includes a new set of sample parameters, based on the latest thinking from the curriculum team, and mapped out in this chart:
  
Screen Shot 2020-12-07 at 8 10 12 PM


Erin007:
One overall thought, which isn't worth addressing right now, but wanted throw out - what happens in the future in RTL languages? We don't fully flip the navigation direction for other parts of our site like clicking through level bubbles, right? And wouldn't need to reverse column orders or anything like that?

public/datasets/erin_candy_preferences.json
```diff
@@ -1,4 +1,5 @@
 {
+  "defaultLabelColumn": "name",
```
Erin007, line 2: I think the default label column in this case is `delicious?`;  the model won't predict a candy's name very well.

public/datasets/titanic.json
```diff
@@ -1,4 +1,5 @@
 {
+  "defaultLabelColumn": "Name",
```
Erin007, line 2: Similar to the candy dataset I think the default label column in this case is `Survived`.

public/index.html
```diff
@@ -11,6 +11,12 @@
       <option disabled selected value="/">Jump to mode...</option>
       <option value="">Default</option>
       <option value="load_foods">Load Foods</option>
+      <option value="level_3_1">3/1</option>
+      <option value="level_3_3">3/3</option>
+      <option value="level_4_1">4/1</option>
+      <option value="level_4_3">4/3</option>
+      <option value="level_5_3">5/3</option>
+      <option value="level_7_6">7/6</option>
     </select>
```
Erin007, line 20: In follow up, let's make this more descriptive rather than tied to a lesson sequence that's still in flux.

src/App.js
```diff
@@ -10,28 +10,161 @@ import Results from "./UIComponents/Results";
...
+                onClick={() => setCurrentPanel(panel.id)}
+              >
+                {panel.label}
+              </div>
+            );
+          })}
+      </div>
+    );
+  }
+}
```
Erin007, line 65: Right now we have 1 component per file; is there an advantage to putting multiple in this parent file?

breville: Just a slight personal preference for not creating new files in the case where they remain small, and have no external consumers.

src/redux.js
```diff
@@ -173,154 +173,166 @@ const initialState = {
...
       ...state,
-      jsonfile: action.jsonfile
+      jsonfile: action.jsonfile,
     };
   }
   if (action.type === SET_IMPORTED_DATA) {
     return {
       ...state,
-      data: action.data
+      data: action.data,
```
Erin007, line 207: what's up with all the trailing commas? does your editor auto add them?

src/redux.js
```diff
@@ -608,3 +644,101 @@ export function getTrainedModelDataToSave(state) {
...
+
+export function getPanels(state) {
+  return panelList
+    .filter(panel => {
+      return isPanelVisible(panel.id);
+    })
+    .map(panel => {
+      return { ...panel, enabled: isPanelEnabled(state, panel.id) };
+    });
+}
```
Erin007, line 744: Yay! This feels much better. Thanks for sticking with it!

Erin007 (approved):

src/App.js
```diff
@@ -10,28 +10,161 @@ import Results from "./UIComponents/Results";
...
+                <FontAwesomeIcon icon={faSquare} />
+                &nbsp;
+                {validationMessages[key].errorString}{" "}
+              </p>
+            );
+          })}
+      </div>
+    );
+  }
+}
```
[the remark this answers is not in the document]
breville: Yes, I think there are probably some big content passes yet to come.

src/redux.js
```diff
@@ -608,3 +644,110 @@ export function getTrainedModelDataToSave(state) {
...
+  { id: "columnInspector", label: "Columns" },
+  { id: "selectTrainer", label: "Trainer" },
+  { id: "trainModel", label: "Train" },
+  { id: "results", label: "Results" },
+  { id: "predict", label: "Predict" },
+  { id: "saveModel", label: "Save" }
+];
+
+export function getPanelEnabled(panel) {
+  const state = store.getState();
```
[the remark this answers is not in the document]
breville: Great point.  The refactor to use the [selector](https://redux.js.org/recipes/computing-derived-data) approach should now address this.

src/redux.js
```diff
@@ -178,7 +185,8 @@ const initialState = {
   prediction: {},
```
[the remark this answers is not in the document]
breville: It is now gone.

src/redux.js
breville: My local JsPrettier settings somehow got reset; will fix.

PR #50, opened 2021-01-04
Validate that models to save have names 

Erin007:
Encourage the user to name their model if they haven't yet done so and only save if named.


Erin007:
We'll need to think through what visual cue of success the user gets when the model is saved successfully, what happens if the save fails and disabling the save button if not named, but this seemed like a good start for a Monday after break.

breville (approved):

PR #52, opened 2021-01-08
Export to App Lab button and function

Erin007:
Adds a new button "Export to App Lab" to the "save" panel that when clicked will save the trained model and pass the model id as a query param to a newly opened App Lab project. 

Compatible with Code Studio PR [#38468](https://github.com/code-dot-org/code-dot-org/pull/38468)

Screen Shot 2021-01-08 at 11 34 30 AM


Erin007:
Closing in favor of focusing on the import of models from the App Lab side rather than pushing from the AI Lab side.

PR #53, opened 2021-01-17
UI update

breville:
This is a pretty substantial UI update for AI Lab. 

There were a few goals:
- reduce the number of panels/tabs.
- use a previous/next button pair to navigate back and forth, similar to [AI for Oceans](https://github.com/code-dot-org/ml-activities), rather than relying on tabs.
- Make the data itself central to the experience, both for exploration and for setting up the training.
- Connect different elements of the experience through proximity, colour, etc.
- Display a clear message about the student's intent.

Here's a glimpse of the most minimal mode we now have:

Screen Shot 2021-01-18 at 12 26 42 PM

The resulting experience is still a work in progress, and needs a lot of refinement and cleanup, but has the following features:

- Fewer panels/tabs in a number of modes.
- Previous/next buttons now show.  The tabs are still there for now, but are subject to removal later.
- The data columns are now clickable, with a metapanel on the right allowing:
  - setting of the column type
  - a frequency histogram
  - colour-coded buttons to set this column as either the label ("predict this") or a feature ("predict based on this")
- Columns are now highlighted when they are acting as a label or feature, with the appropriate colour-code.
- A large sentence at the top forms a "statement of intent" that the student constructs, in particular expressing the label and features, colour-coded.
- The label and features in the "statement of intent" can also be clicked to bring up popups for adjusting the selected columns.  This duplicates functionality available in the metapanel; we'll see if we want to keep both.
- The "training" panel now just shows for a couple seconds.  This is where we can later put some kind of fun animation.

The styling is intentionally placeholder quality.  Proper styling will come later.

Here is that minimal mode in action:

https://user-images.githubusercontent.com/2205926/104860950-8b6d9580-58e2-11eb-8849-70346dde7190.mov


Erin007:
I think this is definitely moving in the right direction! I know this isn't finished yet, but thought I'd share first impressions: 

We shouldn't let you get into a state where you've selected a column as both a feature and a label. 
Screen Shot 2021-01-20 at 11 23 52 AM

I wanted to be able to click the column name heading to interact with the column. 

Wondering about color system for the selected column once you've indicated it should be a label or a feature - the color designation doesn't update until you click off of the column. 

Do we have a plan to revive the validation messages? Even though the UI is more consolidated/interactive I have less of a clear idea of what I'm supposed to do on each tab. 

Predict based on this column switches to "don't predict based on this column", but "predict this column" doesn't toggle like that. We probably want to pick a consistent interaction pattern?

Graphs will need to be able to handle whatever data is in each of the columns - looks like right now only supports when the columns have the same options? 
Screen Shot 2021-01-20 at 11 30 49 AM

Screen Shot 2021-01-20 at 11 32 42 AM

I'm not convinced the auto jump from Train to Results feels right yet. I'm leaning toward still having a train panel that lets you select the % of data to reserve (and maybe where from the set to select it from) and some more algorithm set up.  I think we'll want a paused learning moment, even if we don't always expose all of the algorithm selection and customization?

The minimal flow still needs to include the ability to save a trained model. 

It wasn't immediately clear that I could click on the mad lib at the top.  

The minimal demo has all categorical columns; what's the plan for continuous? 

Should the results panel have a right arrow going to Predict?

If I play with http://localhost:8080/?mode=level_3_1 or any of the other non-minimal modes, I get stuck on the Trainer panel and can't progress even if I select the necessary options. :(

breville:
@Erin007 Thanks for the excellent feedback and good catches.

Some updates have been committed:

- When using the metapanel, you can no longer select a column as both label and feature.  (If we use the popups, we'll need to build in similar protections.)
- Headers can now be clicked in the data table.
- The data table now uses different styling for columns in six conditions: unselected, unselected label, unselected feature, selected, selected label, selected feature.  The upshot is that you can now see it change style while it's selected.
- There is now a "Don't predict this column" button so that you can deselect the label.
- Fixed the graph.
- There is still a Trainer panel which has the reservation UI and additional algorithm setup.  But I have added a spinner to the Train panel to make it clearer that this is where we'll put a fun animation; after a few seconds, it goes away and a next button appears.
- Results now has a next button that goes to Predict.  I'm still pondering whether there's more significant refactoring of those panels, though. 
- The minimal mode now has the Save panel, which has a previous button that goes back to Predict.  There is also now a stub function to save the model which writes to the debug console, for standalone builds.  (Should this be an async function?)
- I've tried putting some dotted boxes around the clickable text in the statement of intent.  Though must admit that I'm finding it much quicker to use the metapanel.
- Also, the columns are now variable width.

Additional comments:

- Yes, I only commented out the validation message code, which is a special occasion because I assume we'll want to bring a bunch of it back.  I wanted to dial back to a minimal experience first so that we could all play with it and get a feel for where we want this scaffolding to happen.  I think we might be able to reduce the amount of messaging by having UI that restricts forward progress until the user does certain things, but I certainly expect to bring a bunch of that messaging back too.
- I don't repro getting stuck in `?mode=level_3_1`.
- The level author can specify the appropriate trainer for continuous data as well.  (Do we need to allow them to specify the K value as well?)

src/UIComponents/SelectTrainer.jsx
```diff
@@ -54,20 +100,19 @@ class SelectTrainer extends Component {
...
+              <label>
+                <p>What would you like the value of K to be?</p>
+                <input
+                  /* value of input is handled by default */
+                  onChange={this.handleChangeKValue}
+                  type="text"
+                />
+              </label>
+            </div>
+          )}
```
Erin007, line 115: thanks for tidying up this bit!

Erin007 (approved):
There's nothing in the code that's jumping out as particularly problematic.  As long as the original flows are unblocked, I say let's ship this so folks can see how it feels. 🚀

PR #57, opened 2021-01-22
User can select where in the dataset the test data comes from 

Erin007:
Early in the curriculum unit, we'd like to ensure that students are creating the same models given the same datasets.  The only way to do this is to eliminate the randomness with which we've been selecting the test data from the data set. This change adds a dropdown for a user to select "beginning", "middle" or "random" for where the test data comes from. 

Screen Shot 2021-01-22 at 3 48 01 PM

If the user has selected to reserve 0% of the dataset, they will not see this additional dropdown. 

Screen Shot 2021-01-22 at 3 48 09 PM

### DIFF
--- a/src/UIComponents/SelectTrainer.jsx
+++ b/src/UIComponents/SelectTrainer.jsx
@@ -5,17 +5,20 @@ import { connect } from "react-redux";
 import {
   getShowChooseReserve,
   setPercentDataToReserve,
+  setReserveLocation,
   setSelectedTrainer,
   getCompatibleTrainers,
   setKValue
 } from "../redux";
-import { styles, TRAINING_DATA_PERCENTS } from "../constants";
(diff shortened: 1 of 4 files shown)


breville (approved):

breville:
Oh, you could also set up the parameterisation for this, to both set a value and hide it from the UI, similar to how we do it for `hideSelectTrainer`.  Otherwise, I can do it next week.  It would be good to keep the `minimal` example mode up to date with parameters like this which would keep it... minimal. :)

Erin007:
> Oh, you could also set up the parameterisation for this, to both set a value and hide it from the UI, similar to how we do it for `hideSelectTrainer`. Otherwise, I can do it next week. It would be good to keep the `minimal` example mode up to date with parameters like this which would keep it... minimal. :)

It will default to selecting from the beginning of the dataset, and I can add parameterization for this as a follow up.

PR #58, opened 2021-01-27
Convert numbers back to human readable strings in Results table 

Erin007:
For categorical data, we convert each category to a number "behind the scenes" before it can be used by the machine learning algorithms.  We're converting the predictions back to human readable strings for the expected and predicted labels in the Results table, but we weren't converting the feature column data in the results table. Now we are. 

Before: 
Screen Shot 2021-01-26 at 9 23 07 PM

After: 
Screen Shot 2021-01-26 at 9 21 09 PM

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -582,8 +582,13 @@ export function getConvertedAccuracyCheckExamples(state) {
 }
 
 export function getConvertedLabel(state, rawLabel) {
-  if (state.labelColumn) {
-    getConvertedValue(state, rawLabel, state.labelColumn);
+  if (state.labelColumn && !isEmpty(state.featureNumberKey)) {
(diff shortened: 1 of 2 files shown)


Erin007:
@brondle FYI

src/redux.js
```diff
@@ -556,6 +556,31 @@ function isEmpty(object) {
...
+  }
+}
+
+export function getConvertedAccuracyCheckExamples(state) {
+  const convertedAccuracyCheckExamples = [];
+  var example;
+  for (example of state.accuracyCheckExamples) {
+    let convertedAccuracyCheckExample = [];
+    var i;
+    for (i = 0; i < state.selectedFeatures.length; i++) {
```
breville, line 574: nit: can you do `var i = 0` inside the `for`?

breville (approved):

PR #62, opened 2021-01-29
CrossTab UI

breville:
This introduces a kind of `CrossTab` UI control, shown when a label and one or more features are selected.  It has a row for each combination of feature values, percentages for the cells, and rudimentary heatmap colouring.

The idea is that students can spot features, and feature combinations, that have particularly strong predictive power over a label.

Screen Shot 2021-01-29 at 5 38 07 PM
Screen Shot 2021-01-29 at 5 38 32 PM
Screen Shot 2021-01-29 at 5 38 58 PM

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -899,12 +899,51 @@ export function getCurrentColumnIsSelectedFeature(state) {
   return state.selectedFeatures.includes(state.currentColumn);
 }
 
+/* Returns an object with information for the CrossTab UI.
+ *
+ * Here is an example result:
+ *
+ *  {
+ *    results: [
+ *      {
+ *        featureValues: ["1", "1"],
+ *        labelCounts: { yes: 2, no: 1 },
+ *        labelPercents: { yes: 67, no: 33 }
+ *      },
+ *      {
+ *        featureValues: ["0", "0"],
+ *        labelCounts: { yes: 25, no: 42 },
+ *        labelPercents: { yes: 37, no: 63 }
+ *      },
+ *      {
+ *        featureValues: ["1", "0"],
+ *        labelCounts: { yes: 6, no: 5 },
+ *        labelPercents: { yes: 55, no: 45 }
+ *      },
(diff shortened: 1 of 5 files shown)


src/redux.js
```diff
@@ -898,3 +898,124 @@ export function getCurrentColumnIsSelectedLabel(state) {
...
+    for (let labelCount of Object.values(result.labelCounts)) {
+      totalCount += labelCount;
+    }
+    result.labelPercents = {};
+    for (let key of Object.keys(result.labelCounts)) {
+      result.labelPercents[key] = Math.round(
+        (result.labelCounts[key] / totalCount) * 100
+      );
+    }
+  }
```
Erin007, line 987: Have you tried this with a big dataset? I have a general concern about performance in scenarios where we go through a whole column; not blocking for this, just curious.

src/redux.js
```diff
@@ -898,3 +898,124 @@ export function getCurrentColumnIsSelectedLabel(state) {
...
+    for (let labelValue of labelValues) {
+      if (
+        !uniqueLabelValues.find(uniqueLabelValue => {
+          return uniqueLabelValue === labelValue;
+        })
+      ) {
+        uniqueLabelValues.push(labelValue);
+      }
+    }
+  }
```
Erin007, line 1004: There's already a function`getUniqueOptions(state, state.currentColumn)`. Reusable here?

Erin007 (approved):
Thanks for pulling this together so quickly! Code LGTM with a couple questions for consideration. Curious to see how the Curriculum team feels about this direction.

src/redux.js
breville: Good call.  Will use that.

src/redux.js
breville: Good point.  I suspect we have a number of perf bottlenecks and we will need to understand how big the incoming datasets can get, but I haven't yet done that.

PR #63, opened 2021-01-30
Add a close X to the column metapanel

breville:
It deselects the current column and allows the user to see the CrossTab again.  Not sure if this is the right UI scheme but it's at least a bit easier.

#### With column selected:

Screen Shot 2021-01-30 at 5 21 45 PM

#### After pressing X:

Screen Shot 2021-01-30 at 5 21 50 PM


Erin007 (commented):
Looks good!

PR #64, opened 2021-02-02
CrossTab now scrolls rather than overflows

breville:
### before

Screen Shot 2021-02-08 at 12 08 07 PM

### after

Screen Shot 2021-02-08 at 12 08 49 PM


Erin007 (approved):
Worth adding screenshots to show the difference?

PR #65, opened 2021-02-07
Model card: initial work

breville:
Initial work on the **Model Card**, which is what the user fills out when they're going to save a trained model.  

(The model card will soon be shown in App Lab when the user is considering importing that trained model.)

Here it is using column descriptions taken from the JSON metadata that we will provide:
Screen Shot 2021-02-07 at 2 24 31 PM

The saved model data will look like this:

```
{
   "name":"My model name",
   "uses":"It can be used for good.",
   "misuses":"It can be misused for evil",
   "columns":[
      {
         "id":"bar",
         "description":"Whether it is a bar"
      },
      {
         "id":"caramel",
         "description":"Whether it has caramel"
      }
   ],
   "identifySubgroup":true,
   "representSubgroup":true,
   "decisionsLife":false,
   "selectedTrainer":"knnClassify",
   "selectedFeatures":[
      "bar"
   ],
   "featureNumberKey":{
      ...
   },
   "labelColumn":"caramel",
   "summaryStat":{
      "type":"classification",
      "stat":"84.42"
   },
   "trainedModel":{
      ...
   }
}
```

And here it is when the user must manually enter column descriptions because they just uploaded a CSV file:

Screen Shot 2021-02-07 at 2 29 01 PM

The saved model data will look like this:

```
{
   "name":"My second model name",
   "uses":"For good.",
   "misuses":"For evil.",
   "columns":[
      {
         "id":"pluribus",
         "description":"Whether it's pluribilistic"
      },
      {
         "id":"nougat",
         "description":"Whether it's nougatesque"
      }
   ],
   "identifySubgroup":false,
   "representSubgroup":true,
   "decisionsLife":true,
   "selectedTrainer":"knnClassify",
   "selectedFeatures":[
      "pluribus"
   ],
   "featureNumberKey":{
      ...
   },
   "labelColumn":"nougat",
   "summaryStat":{
      "type":"classification",
      "stat":"62.34"
   },
   "trainedModel":{
      ...
   }
}
```

It also demonstrates adding column descriptions to the metadata for one of our datasets.  The description is now shown in the column metadata panel when available.

### DIFF
--- a/src/UIComponents/SaveModel.jsx
+++ b/src/UIComponents/SaveModel.jsx
@@ -2,21 +2,25 @@
 import PropTypes from "prop-types";
 import React, { Component } from "react";
 import { connect } from "react-redux";
-import { setTrainedModelDetails, getTrainedModelDataToSave } from "../redux";
+import {
+  setTrainedModelDetail,
+  getTrainedModelDataToSave,
+  getSelectedColumnDescriptions
+} from "../redux";
+import { styles } from "../constants";
 
 class SaveModel extends Component {
   static propTypes = {
     saveTrainedModel: PropTypes.func,
     trainedModel: PropTypes.object,
-    setTrainedModelDetails: PropTypes.func,
+    setTrainedModelDetail: PropTypes.func,
     trainedModelDetails: PropTypes.object,
-    dataToSave: PropTypes.object
+    dataToSave: PropTypes.object,
+    labelColumn: PropTypes.string
   };
 
-  handleChange = (event, field) => {
-    const trainedModelDetails = this.props.trainedModelDetails;
-    trainedModelDetails[field] = event.target.value;
-    this.props.setTrainedModelDetails(trainedModelDetails);
+  handleChange = (event, field, isColumn) => {
+    this.props.setTrainedModelDetail(field, event.target.value, isColumn);
   };
 
   onClickSave = () => {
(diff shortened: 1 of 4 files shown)


src/UIComponents/SaveModel.jsx
```diff
@@ -25,24 +30,84 @@ class SaveModel extends Component {
...
+    fields.push({
+      id: "represent_subgroup",
+      type: "checkbox",
+      text: "Have we ensured the data has adequate representation of subgroups?"
+    });
+    fields.push({
+      id: "decisions_life",
+      type: "checkbox",
+      text:
+        "Can this model could be used to inform decisions central to human life?"
```
Erin007, line 66: Looks like an extra "could" in this sentence...

src/redux.js
```diff
@@ -752,8 +807,31 @@ export function getEmptyCellDetails(state) {
...
+      });
+    }
+  } else {
+    dataToSave.columns = state.trainedModelDetails.columns;
+  }
+
+  dataToSave.uses = state.trainedModelDetails.uses;
+  dataToSave.misuses = state.trainedModelDetails.misuses;
+
+  dataToSave.identify_subgroup = !!state.trainedModelDetails.identify_subgroup;
```
Erin007, line 830: should these be camelCase or like_this? Mostly wondering if they should be consistent and whether I should clean up the fieldsIMade to match these new_ones?

Erin007 (approved):

src/redux.js
breville: Oh, I think camelCase is better; not sure why I did it this way.  Let me update all instances.

PR #66, opened 2021-02-07
Raise the Previous/Next buttons

breville:
Just temporarily, so that level authors don't have the the Previous button hidden behind the Extra Links box.


Erin007 (approved):
Yay! Thanks. This will be a big help.

PR #71, opened 2021-02-15
Results improvements 

Erin007:
A few improvements to the Results panel: 

1.) There was a bug if you picked to reserve 0% of the training data from the beginning of the dataset testing, because `array.slice(0)` basically copies all elements in the array. Now we are correctly not reserving any testing data if the user selects 0%. 

2.) Furthermore, we now hide the results table if 0% of the training data was reserved, because there is nothing to show in the table. Instead we add an explanatory note. 
Screen Shot 2021-02-15 at 5 30 53 PM

3.) Lastly, we're now stating in the results table which column is selected as the label. 
Screen Shot 2021-02-15 at 6 00 03 PM

### DIFF
--- a/src/UIComponents/Results.jsx
+++ b/src/UIComponents/Results.jsx
@@ -69,8 +69,8 @@ class Results extends Component {
                     {this.props.selectedFeatures.map((feature, index) => {
                       return <th key={index}>{feature}</th>;
                     })}
-                    <th>Expected Label</th>
-                    <th>Predicted Label</th>
+                    <th>Expected Label: {this.props.labelColumn}</th>
+                    <th>Predicted Label: {this.props.labelColumn}</th>
                   </tr>
                 </thead>
                 <tbody>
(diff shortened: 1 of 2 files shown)


src/UIComponents/Results.jsx
```diff
@@ -28,59 +28,73 @@ class Results extends Component {
...
-            {this.props.summaryStat.type === MLTypes.CLASSIFICATION && (
-              <div>
-                <div style={styles.mediumText}>
-                  The calculated accuracy of this model is:
-                </div>
-                <div style={styles.subPanel}>
-                  {this.props.summaryStat.stat}%
+              )}
+            {this.props.summaryStat.type === MLTypes.CLASSIFICATION &&
+              !isNaN(this.props.summaryStat.stat) && (
```
breville, line 51: I wonder if some of this kind of logic belongs in redux rather than in the view..

breville (approved):

PR #76, opened 2021-02-22
More UI work

breville:
Significant UI work which moves us closer to a step-by-step interaction model.  Selecting the label is now a standalone step before selecting the features.  The frequency histogram is shown in the former, while the CrossTab is shown in the latter.  Column type selection, if needed, is now done immediately after import.

Each panel, be it full width, a wide one on the left, or a narrow one on the right, uses common styling to handle a variety of vertical window sizes, with fixed-height elements always shown, while the taller elements have a consistent approach to shrinking and being scrolled.  Flexbox is used for this layout.

A lot of legacy UI support has been removed, including tabs.

Significant work remains, including more comprehensive blocking of progress if the current screen's work is not yet done, better user feedback for such conditions, better instructions in general, fleshed out content for a variety of scenes, scatter plot support for continuous columns, automatic column type detection, and general styling.  There are also bigger proposed design improvements to each scene which we are considering as we approach the endgame for the initial release of this tool.

### Some screenshots

Screen Shot 2021-02-24 at 4 26 36 PM
Screen Shot 2021-02-24 at 4 26 48 PM
Screen Shot 2021-02-24 at 4 26 59 PM
Screen Shot 2021-02-24 at 4 27 15 PM
Screen Shot 2021-02-24 at 4 27 28 PM
Screen Shot 2021-02-24 at 4 27 37 PM
Screen Shot 2021-02-24 at 4 27 44 PM


src/UIComponents/ColumnInspector.jsx
```diff
@@ -77,170 +66,96 @@ class ColumnInspector extends Component {
...
+      currentColumnData && (
+        <div
+          id="column-inspector"
+          style={{ ...styles.panel, ...styles.rightPanel }}
+        >
+          <div style={styles.largeText}>Column Information</div>
+          <form>
+            <div>
+              <label>
+                {true && (
```
Erin007, line 78: Are this bit and the `false &&` below necessary?

src/UIComponents/ColumnInspector.jsx
```diff
@@ -77,170 +66,96 @@ class ColumnInspector extends Component {
...
-                {currentColumnIsSelectedLabel && (
-                  <button
-                    type="button"
-                    onClick={this.removeLabel}
-                    style={styles.dontPredictButton}
-                  >
-                    Don't predict this column
-                  </button>
+              {currentColumnData.dataType === ColumnTypes.CATEGORICAL &&
+                labels.length < 5 && (
```
Erin007, line 113: Maybe a comment or variable name for this magic number :)

src/UIComponents/DataDisplay.jsx
```diff
@@ -86,122 +85,110 @@ class DataDisplay extends Component {
...
+              return <p key={i}>{cellDetails}</p>;
+            })}
+          </div>
+        )}
+        {!this.state.showEmptyCellDetails &&
+          this.props.emptyCellDetails.length > 0 && (
+            <button type="button" onClick={this.toggleEmptyCellDetails}>
+              show empty cell details
+            </button>
+          )*/}
```
Erin007, line 191: Do we want to keep this around as a comment?  May just rely on PR history if we need to revive it?

src/UIComponents/SelectDataset.jsx
```diff
@@ -51,6 +54,9 @@ class SelectDataset extends Component {
       download: true
     });
 
+    // We reset state, but we want to keep hiding the trainer selection.
+    this.props.setSelectedTrainer("knnClassify");
```
Erin007, line 58: How do we know what to see this to?

src/constants.js
```diff
@@ -17,14 +17,23 @@ export const TestDataLocations = {
...
 };
 
+const labelColor = "rgb(254, 96, 3)"; // was 186
+const labelColorSemi = "rgba(254, 96, 3, 0.4)";
+const featureColor = "rgb(75, 155, 213)";
+
 export const colors = {
   feature: "rgb(70, 186, 168)",
   label: "rgb(186, 168, 70)"
 };
```
Erin007, line 28: Ah I see - we both had similar ideas but implemented differently. Let's add standardizing and consolidating colors to our list of clean up as we churn on styles.

Erin007 (approved):
Lots to tidy up, but nothing ship blocking from my perspective. Let's get this out in the world and see how folks feel about it. 

LGETM!

src/UIComponents/SelectDataset.jsx
breville: Good point.  I'll change technique to use what was specified by the optional `mode`.

PR #77, opened 2021-02-22
Extract ResultsTable as a sub-component

Erin007:
Initial improvements to the Results table. 

UI improvement: 
- Features are grouped and colored to match other references to features throughout the flow. 
- Predicted column and Actual column flip in order since A.I. Bot makes its guess, _then_ we check if it's correct or not 
- Clearer column labeling and slightly better spacing 
- Added an X when A.I. Bot's guess is wrong 

Code improvements: 
- Extracted the ResultsTable into it's own component 
- Started a colors constant to track feature and label colors

Before: 
Screen Shot 2021-02-22 at 3 12 53 PM

After: 
Screen Shot 2021-02-22 at 3 20 15 PM

### DIFF
--- a/src/UIComponents/ResultsTable.jsx
+++ b/src/UIComponents/ResultsTable.jsx
@@ -1,0 +1,107 @@
+/* React component to handle displaying test data and A.I. Bot's guesses. */
+import PropTypes from "prop-types";
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import {
+  getConvertedAccuracyCheckExamples,
+  getConvertedLabels,
+  getSummaryStat
+} from "../redux";
+import { styles, colors, MLTypes } from "../constants";
+
+class ResultsTable extends Component {
+  static propTypes = {
+    selectedFeatures: PropTypes.array,
+    labelColumn: PropTypes.string,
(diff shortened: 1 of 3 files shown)


src/UIComponents/ResultsTable.jsx
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  render() {
+    return (
+      <div>
+        <div style={styles.floatLeft}>
+          <span style={styles.largeText}>Features</span>
+          <table>
+            <thead>
+              <tr style={{ backgroundColor: colors.feature }}>
```
breville, line 27: We could also have something in `styles` that has the appropriate `backgroundColor` for a given feature, which will give us a bit more flexibility if we want to, say, change the foreground color for such items at the same time.

breville (approved):

src/UIComponents/ResultsTable.jsx
Erin007: Yeah, I was thinking too that we could add additional colors (like the red and green we use for icons & warnings etc) and then use those references in the styles object. Lots to tidy/consolidate as we move towards more finalized styling!

PR #83, opened 2021-03-02
Default to categorical columns and KNN trainers 

Erin007:
- default to categorical data type if not already set 
- move column data type confirmation to TrainingSettings component, along with extracted slider to reserve data for testing
- skip specify columns panel at the beginning of the flow
-  include column type dropdown in single column data visualization panel 
- put all the "advanced controls" in a TrainingSettingsAdvanced component that's current not used  
- default to KNN and use data type of label column to set classification or regression if selectedTrainer isn't explicitly set 

https://user-images.githubusercontent.com/12300669/109667958-ad437480-7b3e-11eb-9e46-e3ffe78f9b25.mov


src/UIComponents/TrainingSettings.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+  render() {
+    return (
+      <div id="select-trainer" style={styles.panel}>
+        <div>
+          <div style={styles.largeText}>Are you ready to train A.I. Bot?</div>
+          <div style={styles.largeText}>
+            A.I. Bot will predict {this.props.labelColumn} based on{" "}
+            {this.props.selectedFeatures.join(", ")}.
+          </div>
```
breville, line 23: We should just have one `largeText` and then smaller text for the second sentence.

src/UIComponents/TrainingSettings.jsx
```diff
@@ -0,0 +1,35 @@
...
+          <PercentReserveSlider />
+          <ColumnTypeSelector />
```
breville, line 25: If we have loaded metadata, I don't think we should show this here.

src/UIComponents/ColumnInspector.jsx
```diff
@@ -77,9 +77,21 @@ class ColumnInspector extends Component {
...
+                  value={currentColumnData.dataType}
+                >
+                  {Object.values(ColumnTypes).map((option, index) => {
+                    return (
+                      <option key={index} value={option}>
+                        {option}
+                      </option>
+                    );
+                  })}
+                </select>
```
breville, line 94: If we have preloaded metadata, I think we should just show the data type here without allowing the user to change it.

src/App.js
```diff
@@ -1,13 +1,11 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import SelectDataset from "./UIComponents/SelectDataset";
-import SpecifyColumns from "./UIComponents/SpecifyColumns";
-import SpecifyColumnsInfo from "./UIComponents/SpecifyColumnsInfo";
```
breville, line 5: Should we delete these files too, then?

src/UIComponents/ColumnInspector.jsx
```diff
@@ -136,7 +153,8 @@ class ColumnInspector extends Component {
 export default connect(
   state => ({
     currentColumnData: getCurrentColumnData(state),
-    rangesByColumn: getRangesByColumn(state)
+    rangesByColumn: getRangesByColumn(state),
+    hideSpecifyColumns: state.mode.hideSpecifyColumns
```
breville, line 157: This hits an exception when `mode` is `null` which happens in our default standalone invocation.

breville: (We guard against it [here](https://github.com/code-dot-org/ml-playground/pull/83/files#diff-6e02291d2b49a002990985f9cb58187da3ad794009887859ccce802714a852ddL32), for example.)

src/App.js
Erin007: done!

src/UIComponents/ColumnInspector.jsx
Erin007: done! Thanks for the tip.

breville (approved):

PR #86, opened 2021-03-03
add much dataset

brondle:
Added new datasets for the library with JSON metadata, images coming soon!


Erin007 (approved):

dancodedotorg:
:wave: A request: before this gets merged, can we also update the `datasetManifest.js` file to reference these new datasets too? Otherwise they won't show up in the tool. (since this is from Brent's fork, I don't think I can do this myself =/)

brondle:
> > 👋 A request: before this gets merged, can we also update the `datasetManifest.js` file to reference these new datasets too? Otherwise they won't show up in the tool. (since this is from Brent's fork, I don't think I can do this myself =/)
> 
> @brondle Is that something you can pick up?

Sure thing!

dancodedotorg:
@brondle FYI - I snuck in a PR before this one (https://github.com/code-dot-org/ml-playground/pull/92) which also updated the datasetManifest.js file, so you may have to pull my changes first before adding yours.

brondle:
@dancodedotorg @Erin007 added a few more sets and updated the manifest. Erin, I just got the 007 joke in your username 🤦‍♂️

src/datasetManifest.js
```diff
@@ -251,7 +249,85 @@ export const allDatasets = [
     path: "datasets/medical_priority.csv",
     metadataPath: "datasets/medical_priority.json",
     imagePath: "datasets/temp.jpg"
+  },
+  {
+    id: "nutrition",
```
dancodedotorg, line 254: @brondle beep boop - looks like nutrition shows up twice? Here and line 261?

brondle: whoops should be fixed

PR #93, opened 2021-03-05
Display warning when uploaded dataset has >20k rows

made-line:
I added an alert in the DataCard component. It warns users when their dataset contains more than 20,000 rows. 

Edited to add screenshot: 
dataset-warning

### DIFF
--- a/src/UIComponents/DataCard.jsx
+++ b/src/UIComponents/DataCard.jsx
@@ -17,7 +17,7 @@ class DataCard extends Component {
 
     const card = metadata && metadata.card;
 
-    const dataLengthLimit = 20000
+    const dataLengthLimit = 20000;
     if (dataLength > dataLengthLimit) {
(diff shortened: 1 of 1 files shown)


src/UIComponents/DataCard.jsx
```diff
@@ -17,6 +17,11 @@ class DataCard extends Component {
 
     const card = metadata && metadata.card;
 
+    const dataLengthLimit = 20000
```
Erin007, line 20: nit:  ; at the end of this line.

Erin007:
Can you include a screen shot of the warning, please? Thanks!

Erin007 (approved):

breville (approved):
Nice work!  

As discussed, we should probably ponder a consistent, long-term messaging solution for all of our warnings, but we don't have it yet.

PR #95, opened 2021-03-06
Save button on final panel saves then continues

breville:
Saving the trained model is now done with a regular navigation button labeled "Save".  Once successfully saved, the browser navigates to the next level.

If the `SaveModel` panel is hidden, then the results panel has a button labeled "Continue" that navigates to the next level.

Because the navigation buttons live outside the scope of a single panel, some of the logic has been hoisted upwards.

Screen Shot 2021-03-07 at 12 36 09 AM


Erin007:
I think this is the right move for this context, but noting I wonder about the future standalone experience. For example, would clicking save take the user to a new app lab project, then?

breville:
@Erin007 It's a great question.  I don't know the answer yet, but I think it will depend on how we present the standalone experience, and could require a variation.

Erin007 (approved):

PR #97, opened 2021-03-08
Dynamic instructions

breville:
Support for a new type of Dynamic Instructions, as provided by the main repo's `TopInstructions`.  The AI Lab app sends a key back to its host indicating which instruction should be shown at that time.  Currently, the keys are the same as the `currentPanel` values, giving a unique instruction per panel.

The standalone app just shows that key value at the top.

See main repo work in https://github.com/code-dot-org/code-dot-org/pull/39384.

Screen Shot 2021-03-16 at 6 58 46 AM


src/App.js
```diff
@@ -193,7 +193,7 @@ class App extends Component {
 
         {currentPanel === "results" && (
           <BodyContainer>
-            <ContainerLeft>
+              <ContainerLeft>
```
Erin007, line 196: this indentation doesn't look intentional?

Erin007 (approved):

src/App.js
breville: Good catch, thanks!

PR #98, opened 2021-03-09
Update strings

made-line:
**Select Data - before**
select_data_before

**Select Data - after**
select_data_after

**Select Label - before**
select_label_before

**Select Label - after**
(Edited to add)
select_label_fix

select_label_after_numerical

**Training Preparation - before**
training_preparation_before
training_preparation_button_before

**Training Preparation - after**
training_preparation_after
training_preparation_button_after

**Results - before**
results_before

**Results - after**
results_after

**Tests - before**
test_before

**Tests - after**
test_after

**Export - before**
export_before

**Export - after**
save_model_fix

### DIFF
--- a/src/UIComponents/ColumnInspector.jsx
+++ b/src/UIComponents/ColumnInspector.jsx
@@ -106,7 +106,8 @@ class ColumnInspector extends Component {
           <form>
             <div>
               <label>
-                <div>Data type:</div>
+                <div>{currentColumnData.id}</div>
+                <div>Data Type:</div>
                 {this.props.hideSpecifyColumns && (
                   <div> {currentColumnData.dataType} </div>
                 )}
--- a/src/UIComponents/ColumnInspector.jsx
+++ b/src/UIComponents/ColumnInspector.jsx
@@ -106,7 +106,7 @@ class ColumnInspector extends Component {
           <form>
             <div>
               <label>
-                <div>{currentColumnData.id}</div>
+                <div>Data type:</div>
                 {this.props.hideSpecifyColumns && (
                   <div> {currentColumnData.dataType} </div>
                 )}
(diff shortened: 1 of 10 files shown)


src/UIComponents/ColumnInspector.jsx
```diff
@@ -106,7 +106,7 @@ class ColumnInspector extends Component {
           <form>
             <div>
               <label>
-                <div>{currentColumnData.id}</div>
```
Erin007, line 109: We need to keep this **and** add "Data Type:" below it.

src/UIComponents/SaveModel.jsx
```diff
@@ -22,21 +22,26 @@ class SaveModel extends Component {
...
+        text: "Description for: " + columnDescription.id,
         answer: columnDescription.description
       });
     }
 
-    fields.push({ id: "potentialUses", text: "How can this model be used?" });
+    fields.push({
+      id: "potentialUses",
+      text: "How can this model be used?",
+      placeholder: "Write a brief description"
```
Erin007, line 39: teeny tiny nit: do we want punctuation at the end of the placeholder?

src/UIComponents/ColumnInspector.jsx
made-line: Added it back in!

src/UIComponents/SaveModel.jsx
made-line: Good catch! I added a period and updated the screenshots :)

src/constants.js
```diff
@@ -1,6 +1,6 @@
 export const ColumnTypes = {
   CATEGORICAL: "categorical",
-  CONTINUOUS: "continuous"
+  CONTINUOUS: "numerical"
```
breville, line 3: This impacted the ability to use some of our datasets, since their metadata still refers to `"continuous"`.  Was it intentional?

Erin007: We wanted the string "continuous" to change to "numerical", but not to impact the logic or datasets. This is totally on me for not being clear. We can either change everything to numerical OR keep the strings and the logic/datasets separate. I vote for the latter.

Erin007: Good catch - sorry I missed this initially! #110

PR #99, opened 2021-03-09
Prevent save if model name is too long

Erin007:
If a user tried to save a model with a name longer than 256 characters, we hit `Mysql2::Error: Data too long for column 'name'` ([HoneyBadger](https://app.honeybadger.io/projects/3240/faults/77689344#notice-trace)) because the name column on the UserMLModels table has a maximum string length.  To prevent this, we now set a maximum length for the name input box that cuts off the entered text if a user tries to add a really long name.


breville:
How about we just restrict the input field to fewer than 256 characters using React?  I think it would be simpler, and makes for one less string to worry about showing & maintaining.

Erin007:
> How about we just restrict the input field to fewer than 256 characters using React? I think it would be simpler, and makes for one less string to worry about showing & maintaining.

Good idea! Today I learned about `maxLength` for input fields 😄

src/constants.js
```diff
@@ -21,6 +21,8 @@ export const TestDataLocations = {
   RANDOM: "random"
 };
 
+export const ModelNameMaxLength = 256;
```
breville, line 24: Should we make it smaller just because anything above something like (say) 50, feels like it might be a mistake?

src/UIComponents/SaveModel.jsx
```diff
@@ -83,6 +83,7 @@ class SaveModel extends Component {
                         onChange={event =>
                           this.handleChange(event, field.id, field.isColumn)
                         }
+                        maxLength={field.id === "name" ? ModelNameMaxLength : ""}
```
breville, line 86: Shoudl we also enforce length limits on the other fields?

Erin007: The other fields are stored as metadata, which is a text column in the database and has a much higher limit of 65,535 characters, so I think it's ok not to set a max length for the other fields.

breville (approved):

src/UIComponents/SaveModel.jsx
breville: I feel like we've occasionally seen people paste an entire work of Shakespeare as a field in a form (for real) so it might be convenient for us to limit these to sane sizes, but we can ponder for a future work item.

PR #106, opened 2021-03-12
Remove the result predicted when a user selects different data

made-line:
[Trello task ](https://trello.com/c/I45sSEUf/113-when-user-selects-different-data-in-test-tool-remove-the-result-predicted)

https://user-images.githubusercontent.com/40412372/110993616-6159b200-832c-11eb-8cce-9b68bc5c0780.mov

### DIFF
--- a/src/UIComponents/Predict.jsx
+++ b/src/UIComponents/Predict.jsx
@@ -9,8 +9,7 @@ import {
   getSelectedCategoricalFeatures,
   getUniqueOptionsByColumn,
   getConvertedPredictedLabel,
-  getPredictAvailable,
-  setPrediction
+  getPredictAvailable
 } from "../redux";
 import { styles } from "../constants";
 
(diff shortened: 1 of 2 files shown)


src/UIComponents/Predict.jsx
```diff
@@ -23,13 +24,15 @@ class Predict extends Component {
...
-    getPredictAvailable: PropTypes.bool
+    getPredictAvailable: PropTypes.bool,
+    setPrediction: PropTypes.func
   };
 
   handleChange = (event, feature) => {
     const testData = this.props.testData;
     testData[feature] = event.target.value;
     this.props.setTestData(testData);
+    this.props.setPrediction({});
```
breville, line 35: One alternative approach would be to have the `SET_TEST_DATA` reducer also clear the `prediction`.  It might be a bit more robust by centralising the business logic in the reducer rather than out here in the UI.

made-line: Thank you for the feedback! I just pushed an update.

breville (approved):

Erin007 (approved):

PR #107, opened 2021-03-15
Save form updates

Erin007:
This PR includes a few incremental updates to move the current state of the Save Panel closure to the proposed design: 

1.) Collapsible section for column descriptions, including column count and whether each column is a feature or label 
2.) Remove checkboxes for misuses and instead include those ideas in the misuses prompt 
3.) Adjust width of uses/misuses input boxes to encourage more detailed user input 

BEFORE: 

https://user-images.githubusercontent.com/12300669/111186095-0a9ce400-8589-11eb-9d37-023d11cda2d4.mov

AFTER: 

https://user-images.githubusercontent.com/12300669/111185677-9b26f480-8588-11eb-94f4-20eafab44a55.mov

DESIGN COMP: 

Screen Shot 2021-03-15 at 12 14 11 PM

PR #109, opened 2021-03-15
Latest data display UI

breville:
This is the latest data display UI which combines a few of our recent explorations into one concept.  (It's known informally as "old two screens but with popups and the phrase builder".)

As in some earlier work, selecting a label is done independently before selecting features.  The `PhraseBuilder` panel on the right is being used to make the primary action more obvious, rather than requiring clicks on columns.  Clicks on columns are now used for the secondary action, which is exploring graphs.  When selecting a label, the graphs are per-column (bar chart for categorical, and right now just some information about numerical).  When selecting features, the graphs show how that column relates to the previously-selected label (`CrossTab` for categorical, scatter plot for numerical).

https://user-images.githubusercontent.com/2205926/111222114-a1bd6800-8598-11eb-8dd4-3d6fab6dee4c.mov

https://user-images.githubusercontent.com/2205926/111222145-aa15a300-8598-11eb-87ef-d8988d4b1aa8.mov


public/datasets/foods.json
```diff
@@ -31,11 +31,11 @@
   	},
   	{
   		"id": "fruit_day",
-  		"type": "continuous"
+  		"type": "numerical"
   	},
   	{
   		"id": "veggies_day",
-  		"type": "continuous"
```
Erin007, line 38: do we need to do this for all the other datasets too?

src/UIComponents/ColumnInspector.jsx
```diff
@@ -13,6 +13,10 @@ import {
 } from "../redux";
 import { ColumnTypes, styles } from "../constants.js";
 import { Bar } from "react-chartjs-2";
+import ScatterPlot from "./ScatterPlot";
+import CrossTab from "./CrossTab";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faTimes } from "@fortawesome/free-solid-svg-icons";
```
Erin007, line 19: @made-line this is what I was talking about regarding icons in #107  - I used a shortcut where the icon only shows up in the context of code studio, but should arguably switch to this format so it shows up in standalone too.

src/UIComponents/DataDisplay.jsx
```diff
@@ -64,29 +83,46 @@ class DataDisplay extends Component {
...
+            {" "}
+            based on{" "}
+            <span style={styles.statementFeature}>
+              {this.props.selectedFeatures.length > 0
+                ? this.props.selectedFeatures.join(", ")
+                : ".."}
             </span>
+            {"."}
+          </span>
+        </div>*/}
```
Erin007, line 111: can we delete this?

src/UIComponents/PhraseBuilder.jsx
```diff
@@ -0,0 +1,143 @@
+/* React component to handle setting datatype for selected columns. */
```
Erin007, line 1: I don't think this is quite what this component handles... maybe "to handle selecting label and features" ?

src/redux.js
```diff
@@ -455,49 +468,38 @@ export default function rootReducer(state = initialState, action) {
...
-          currentColumn: action.currentColumn
-        };
-      } else {
-        return state;
-      }
+    if (state.currentColumn === action.currentColumn) {
+      return {
+        ...state,
+        currentColumn: undefined
+      };
```
Erin007, line 497: this simplification is very satisfying

Erin007 (approved):
I don't see anything ship blocking except the text being cutoff in PhraseBuilder dropdowns in Code Studio.  

Follow ups...  
1.) you shouldn't be able to create a combo graph with the same column on the x and y axis. 2.) if you select a bunch of features make sure we can scroll to the bottom of them 
3.) ensuring that we use max options = 5 for both the label and the selected feature in the bar chart 
4.) checking graphs on the far right column to make sure nothing's cut off (I think this will be a non-issue if we address the item above)

src/UIComponents/DataDisplay.jsx
breville: I will do so.  I think we should be able to show it on subsequent screens though, to remind the user what they are predicting...

breville:
@Erin007 Great catches.

- The user should now be prevented from selecting the label column when in the feature-selection panel.
- `PhraseBuilder` is now scrollable.
- There are three criteria that can make a `CrossTab` large; we now check against the three of them and hide if any are exceeded.  We can adjust this going forward, too.

Erin007 (approved):
Thanks for addressing the follow-ups! LGTM

src/UIComponents/ColumnInspector.jsx
made-line: Gotcha! Thank you :)

PR #110, opened 2021-03-16
Change continuous to numerical in datasets, function names, etc 

Erin007:
Follow-up to fix fall-out from #98. We're relying on a string comparison to check data types for columns set in dataset JSON files. The comparison string was changed from "continuous" to "numerical" for display purposes; this PR updates all the other instances of "continuous" to "numerical" so the type checking will still work.


breville (approved):
Thanks!

made-line:
Thank you!

PR #113, opened 2021-03-16
Results sticky header

made-line:
Before

https://user-images.githubusercontent.com/40412372/111373795-93835080-8659-11eb-9e9e-0ef86bd8d817.mov

After

https://user-images.githubusercontent.com/40412372/111373831-9da54f00-8659-11eb-911c-3fdd02ff3e5e.mov

Updated after

https://user-images.githubusercontent.com/40412372/111381879-89664f80-8663-11eb-9fd9-0a3bf7e8222e.mov

I also made the table style consistent with what's on the data display page. I can revert this back; I just wanted to see what you both think. If we like the changes, I can update the style naming in constants.js and figure out what to do with the grey border.


src/UIComponents/ResultsTable.jsx
```diff
@@ -25,61 +25,65 @@ class ResultsTable extends Component {
...
-          </thead>
-          <tbody>
-            <tr>
-              {this.props.selectedFeatures.map((feature, index) => {
+      <div style={styles.panel}>
+        <div style={styles.tableParent}>
+          <table style={styles.dataDisplayTable}>
+            <thead>
+              <tr>
+                <th colSpan={featureCount} style={{...styles.largeText, ...styles.dataDisplayHeader, ...styles.resultsTableFirstHeader }}>
```
breville, line 33: You can use [Prettier](https://prettier.io/) to get the formatting looking just right.

breville:
You can see a outer set of scrollbars:

Screen Shot 2021-03-17 at 7 48 05 AM

Remove these lines from `Results.jsx` to get rid of them:
```
          <div style={styles.scrollableContentsTinted}>
            <div style={styles.scrollingContents}>
```

And then those scrollbars will go away:

Screen Shot 2021-03-17 at 7 46 42 AM

src/UIComponents/ResultsTable.jsx
made-line: Thanks for the tip! I pushed changes based off your comments and updated some of the style names.

breville (approved):

PR #116, opened 2021-03-18
Log of trained models and their accuracy 

Erin007:
- [x] log to local storage each time a model is trained 
- [ ] pull from that local storage entry and use the data to generate a table of runs 
- [ ] table row component shows features, label, accuracy and save button for model 
- [ ] display new record of runs table on results screen (figure out how it works w/ current content)

### DIFF
--- a/src/UIComponents/Results.jsx
+++ b/src/UIComponents/Results.jsx
@@ -68,8 +68,10 @@ class Results extends Component {
           !isNaN(this.props.summaryStat.stat) && <ResultsTable />
         }
 
-        <ResultsSummaryTable />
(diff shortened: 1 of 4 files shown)


Erin007:
Looping you in early to see if this approach feels good. It crams quite a lot of stuff into localStorage, including the trained model, but I couldn't think of another place to put it so that the data would be in the right place when/if the user does decide to save a particular trained model from a previous loop through the flow. 

Screen Shot 2021-03-17 at 11 20 41 PM

Erin007:
Closing in favor of #159

PR #117, opened 2021-03-18
Optional parameter to hide column clicking

breville:
This adds a new optional parameter called `hideColumnClicking` which will disable column hovering and clicking, which means no popups with column metadata.  We suspect this might be necessary in earlier levels.

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -496,18 +496,23 @@ export default function rootReducer(state = initialState, action) {
   }
   if (action.type === SET_CURRENT_COLUMN) {
     if (!getShowColumnClicking(state)) {
-      // Do nothing.
+      // If no column clicking, do nothing.
+      return state;
+    }
+    if (
+      state.currentPanel === "dataDisplayFeatures" &&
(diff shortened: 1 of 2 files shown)


src/redux.js
```diff
@@ -487,13 +487,17 @@ export default function rootReducer(state = initialState, action) {
...
+      return {
+        ...state,
+        highlightColumn: action.highlightColumn
+      };
+    }
   }
   if (action.type === SET_CURRENT_COLUMN) {
-    if (state.currentColumn === action.currentColumn) {
+    if (!getShowColumnClicking(state)) {
+      // Do nothing.
```
Erin007, line 499: Do we need to `return {...state}` here?

breville: You're right that this was a bit confusing.  I was trying to have all no-ops fall through to the `return state` at the end of the function, but the logic was getting hard to read.  I'm going to rearrange.

Erin007: Thanks for clarifying & for the comments!

Erin007 (approved):

PR #120, opened 2021-03-19
Training animation with A.I. bot

breville:
This updates the training scene to have an animation with the A.I. bot, inspired by the training scene in [AI for Oceans](https://code.org/oceans).

https://user-images.githubusercontent.com/2205926/112196939-c46b0480-8bc8-11eb-903f-6a12959041a1.mov


src/UIComponents/TrainModel.jsx
```diff
@@ -23,20 +34,103 @@ class TrainModel extends Component {
...
+
+  updateAnimation = () => {
+    if (this.state.frame === 15) {
+      this.setState({ headOpen: true });
+    }
+
+    this.setState({ frame: this.state.frame + 1 });
+  };
+
+  componentWillUnmount = () => {
```
made-line, line 54: First time seeing componentWillUnmount out in the wild. Very nice!

src/redux.js
```diff
@@ -1161,7 +1167,11 @@ export function getPanelButtons(state) {
       next = { panel: "results", text: "Continue" };
     }
   } else if (state.currentPanel === "results") {
-    prev = { panel: "dataDisplayLabel", text: "Back" };
+    prev = isPanelEnabled(state, "dataDisplayLabel")
+      ? { panel: "dataDisplayLabel", text: "Back" }
+      : isPanelEnabled(state, "dataDisplayFeatures")
+      ? { panel: "dataDisplayFeatures", text: "Back" }
+      : null;
```
made-line, line 1174: Interesting use of nested ternary operators! I'm trying to understand why this was added. Does this go hand in hand with the changes in src/index.js?

made-line (approved):
Nice work! I like how the animation can be dynamically resized.

src/redux.js
breville: Good catch.  This was an opportunistic fix for the `hideSelectLabel` mode; I noticed when using the tool to build this new feature that it was still possible to get back to the label scene when it should not have been possible.

breville:
@made-line Thanks for checking resize.  I must admit to not doing anything too explicit to support resizing, but also noticed that it seemed to work pretty well for a variety of screen widths and heights, which is always important for our UI.

src/UIComponents/DataTable.jsx
```diff
@@ -0,0 +1,197 @@
+/* React component to handle displaying imported data. */
```
Erin007, line 1: Hooray for smaller, extracted components!

Erin007 (approved):
Did Mark give a thumbs up too?

breville:
We have a new background image from @markabarrett so I take that as approval.

PR #122, opened 2021-03-22
Added a range of numerical values for testing the model

made-line:
I added in a range of values for numerical features so users know the scale when testing a model. Also updated the spacing in between continuous features. 

Before: 
Screen Shot 2021-03-22 at 4 41 00 PM

Screen Shot 2021-03-23 at 4 18 04 PM

After: 
Screen Shot 2021-03-29 at 4 49 10 PM

Changes to spacing in the context of Code Studio
Before:
Screen Shot 2021-03-29 at 4 25 17 PM

After: 
Screen Shot 2021-03-29 at 5 18 34 PM

### DIFF
--- a/src/UIComponents/Predict.jsx
+++ b/src/UIComponents/Predict.jsx
@@ -9,7 +9,8 @@ import {
   getSelectedCategoricalFeatures,
   getUniqueOptionsByColumn,
   getConvertedPredictedLabel,
-  getPredictAvailable
+  getPredictAvailable,
+  getRangesByColumn
 } from "../redux";
 import { styles } from "../constants";
 
@@ -23,7 +24,8 @@ class Predict extends Component {
     setTestData: PropTypes.func.isRequired,
     predictedLabel: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     confidence: PropTypes.number,
-    getPredictAvailable: PropTypes.bool
+    getPredictAvailable: PropTypes.bool,
+    rangesByColumn: PropTypes.object
   };
 
   handleChange = (event, feature) => {
@@ -47,7 +49,9 @@ class Predict extends Component {
                 return (
                   <div key={index}>
                     <label>
-                      {feature}: &nbsp;
+                      {feature} [{this.props.rangesByColumn[feature].min},{" "}
+                      {this.props.rangesByColumn[feature].max}
+                      ]: &nbsp;
                       <input
                         type="number"
                         onChange={event => this.handleChange(event, feature)}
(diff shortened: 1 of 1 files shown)


dancodedotorg (approved):

breville:
Do students understand this nomenclature? 

We show range in the popup for a numerical column but we just show "min", "max", and "range".  Should we be consistent here?

Finally, should we truncate to maybe 1 or 2 decimal places to make it more readable?

Erin007:
> We show range in the popup for a numerical column but we just show "min", "max", and "range". Should we be consistent here?

+1 to making it more clear what these numbers are. Maybe something like "Age (min: 1, max: 84)". Currently this looks like an array, which might be confusing to students.  I don't think we need range in this scenario.

made-line:
Thank you both! I made changes accordingly and updated the "After" screenshot :)

src/UIComponents/Predict.jsx
```diff
@@ -47,7 +49,10 @@ class Predict extends Component {
                 return (
                   <div key={index}>
                     <label>
-                      {feature}: &nbsp;
+                      {feature} (min:{" "}
+                      {+this.props.rangesByColumn[feature].min.toFixed(2)}, max:{" "}
+                      {+this.props.rangesByColumn[feature].max.toFixed(2)}
```
breville, line 54: Could we hoist these long expressions into a couple local variables in the `render` function, and then use string interpolation to make this expression more compact?  

Also are the `+`s necessary?

breville:
Just curious about the screenshots: why are the first two items touching, while the third is separate?  While we are in here, could we tidy up that spacing, especially between the first two?

src/UIComponents/Predict.jsx
made-line: Good point! If I put the variables in the render function, it loses the context of feature from selectedNumericalFeatures.map((feature, index). Do we want to do this or is the change I just pushed okay?

made-line:
Nice catch. I just added the same padding that's applied to categorical
features. I also added a new screenshot to the PR. Let me know if you think
the padding should be adjusted.

src/UIComponents/Predict.jsx
breville: Oh I see.  There are a few ways to factor this, such as having another React component, or a sub-render function, or a new data structure entirely (such as we do with `getCrossTabData`) but I think this is okay as-is.

breville (approved):

src/UIComponents/Predict.jsx
made-line: Those are good things for me to think about going forward. Thank you!

PR #121, opened 2021-03-22
Adding dataset images

moneppo:

PR #124, opened 2021-03-23
Dataset image metadata

breville:
Followup to https://github.com/code-dot-org/ml-playground/pull/121


made-line (approved):
Looks good. Do we intend to have both full and reduced versions of certain datasets (e.g., full and partial census data)?

src/datasetManifest.js
```diff
@@ -135,7 +135,7 @@ export const allDatasets = [
   {
     id: "zoo", name: "Zoo", path: "datasets/zoo.csv",
```
Erin007, line 136: nit: can we put these on separate lines while you're in this file?

Erin007 (approved):
one tiny nit, but LGTM!

PR #129, opened 2021-03-29
Fix jump to features

breville:
The "Continue" button is now shown after a dataset is chosen and `hideSelectLabel` is used; it will jump directly to features.

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -1139,6 +1139,8 @@ export function getPanelButtons(state) {
     prev = null;
     next = isPanelEnabled(state, "dataDisplayLabel")
       ? { panel: "dataDisplayLabel", text: "Continue" }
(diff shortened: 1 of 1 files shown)


made-line (approved):
Is this meant to handle a scenario in levelbuilder where a dataset and label is preselected for you?

breville:
@made-line This handles the scenario in which the label is preselected (using parameterisation) but the dataset isn't.  It fixes the select-dataset scene so that it now shows the continue button once the user clicks on a dataset.

made-line:
Thank you for the clarification!

breville:
@Erin007 I think this would just count as incompatible parameters and I'd hope nobody combines them.  We could add a guard against this, but I'd hope we'd catch it if ever used in curriculum pretty quickly.

PR #130, opened 2021-03-30
Styling updates

breville:
Some general styling updates have been made.

Also, in particular:

- The statement has been restored to DataDisplay and is also shown in TrainModel, Results, and SaveModel.
- SelectDataset now shows mouse hover highlighting.
- DataDisplay now styles its columns for highlight (due to mouse hover) & selected, and column headers for label & feature.
- The PhraseBuilder now uses the terms label & feature.
- TrainingSettings is now hidden if both `hideSpecifyColumns` & `hideChooseReserve` parameters are set.
- Going back from Results goes to feature selection, while going back from feature selection clears features and goes to label selection, making for a symmetrical forward & back navigation.
- The prediction & actual columns have been swapped so that actual results are closest to the source data.
- The predict button now sizes for its content & goes to half-opacity when disabled.
- Arrows are removed from the navigation buttons.
- The buttons no longer slightly overlap the UI in the standalone version.

### screenshots

Screen Shot 2021-03-31 at 8 46 10 AM

Screen Shot 2021-03-31 at 8 44 00 AM

Screen Shot 2021-03-31 at 8 44 19 AM

Screen Shot 2021-03-31 at 8 44 33 AM

Screen Shot 2021-03-31 at 8 44 50 AM

Screen Shot 2021-03-31 at 8 45 02 AM

### DIFF
--- a/src/UIComponents/Statement.jsx
+++ b/src/UIComponents/Statement.jsx
@@ -1,0 +1,52 @@
+/* React component to display a statement about our model. */
+import PropTypes from "prop-types";
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import { styles } from "../constants";
+
+class Statement extends Component {
+  static propTypes = {
+    data: PropTypes.array,
+    labelColumn: PropTypes.string,
+    selectedFeatures: PropTypes.array
+  };
+
+  render() {
+    const {
+      data,
+      labelColumn,
+      selectedFeatures
+    } = this.props;
+
+    if (data.length === 0) {
+      return null;
+    }
+
+    return (
+      <div style={styles.statement} id="statement">
+        Predict{" "}
+        <span style={styles.statementLabel}>
+          {labelColumn || "..."}
+        </span>
+        <span>
+          {" "}
+          based on{" "}
+          <span style={styles.statementFeature}>
+            {selectedFeatures.length > 0
+              ? selectedFeatures.join(", ")
+              : ".."}
+          </span>
+          {"."}
+        </span>
+      </div>
+    );
+  }
+}
+
+export default connect(
+  state => ({
+    data: state.data,
+    labelColumn: state.labelColumn,
+    selectedFeatures: state.selectedFeatures
+  })
+)(Statement);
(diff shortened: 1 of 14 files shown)


src/UIComponents/Statement.jsx
```diff
@@ -0,0 +1,52 @@
+/* React component to display a statement about our model. */
```
Erin007, line 1: Hooray for extracted components!

src/constants.js
```diff
@@ -225,60 +238,65 @@ export const styles = {
...
+  dataDisplayHeader: {
+    paddingLeft: 20,
+    textAlign: "right",
+    position: "sticky",
+    top: 0,
+    backgroundColor: "white",
+    color: "#4d575f",
+    borderStyle: "solid",
+    borderWidth: 1,
+    borderColor: "white", // #c6cacd",
```
Erin007, line 250: tiny nit: Do we need these little comments here and line 171? I'm not sure when we're using words, when hex and when rgb and why.

Erin007:
Screen Shot 2021-03-30 at 4 30 43 PM
While you're doing a style pass is it worth making "Features" the same size as "Actual" and "A.I. Prediction" and left aligning it?

Erin007:
Screen Shot 2021-03-30 at 4 33 07 PM

Is partial census data also selected or previously selected? Not sure why it also looks like it has a blue outline.

Erin007 (approved):
A few little things to consider, but nothing ship blocking.

src/constants.js
breville: Oh, I will remove this detritus.

breville:
> While you're doing a style pass is it worth making "Features" the same size as "Actual" and "A.I. Prediction" and left aligning it?

Funnily, I thought it might have been deliberate for some reason.  Will fix.

made-line (approved):
Looks good! I like the symmetrical forward and back navigation. 

Do we want to keep the different styling for the data table (gray background) and results table (white background)?

breville:
> Do we want to keep the different styling for the data table (gray background) and results table (white background)?

I was wondering that too.  I think we might do some more work on Results soon, and can consider that then.  I think it might be worth differentiating them, though not entirely clear how, yet.

PR #131, opened 2021-03-30
Hide Full Census and Raw Global Happiness datasets

made-line:
Certain datasets have reduced and full versions of them. We're temporarily hiding the full versions per Mike's recommendation. I hid the Full Census and Raw Global Happiness datasets.

### DIFF
--- a/src/datasetManifest.js
+++ b/src/datasetManifest.js
@@ -48,13 +48,13 @@ export const allDatasets = [
     metadataPath: "datasets/billionaires.json",
     imagePath: "datasets/billionaires.jpg"
   },
-  {
-    id: "census_knn",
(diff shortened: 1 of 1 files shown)


Erin007:
How temporary will this be? I'm not 100% sold on the idea that we'll ever want to pull these full datasets back in to the tool, which makes me think that it'd be slightly cleaner to fully delete these datasets from the manifest and then have a comment at the top of the file that says we're only using datasets with 20,000 rows or less. We can always resurrect via GitHub history and I'd rather not have commented code floating around at this stage in the process.

made-line:
New PR: https://github.com/code-dot-org/ml-playground/pull/137

PR #138, opened 2021-04-07
Update dropdown test scenarios

dancodedotorg:
Updated the test scenarios from the outside-of-levelbuilder version of AI Lab to match the inside-of-levelbuilder parameterized levels - see screenshot below (sorry it's so big - I don't know how to quickly resize images in this flavor of markdown).

Each place in the dropdown matches a level used in the curriculum. Now, if we need to test changes that might affect paramaterized levels, these dropdowns should accurately reflect all of the level types in the curriculum. 

I put this together after the [Dataset Cleanup PR](https://github.com/code-dot-org/ml-playground/pull/137) - I wanted to verify the changes didn't have an unexpected impact on parameterized levels (they don't - everything looks great) and realized it made sense to update these now that the curriculum is more finalized.

Testing story: verified that it works correctly on my local machine.

Capture (Small)

PR #139, opened 2021-04-07
Delete SVM

Erin007:
In collaboration with the Curriculum team we've determined we can meet all of the learning goals for these tools without continuing to support the SVM algorithm. 

This PR includes changes to remove references to SVM including: 
- removing the ml-svm package 
- removing the UI to select between SVM and KNN 
- removing SVM specific test modes in the standalone tool 
- removing code to validate that trainers are compatible with selected labels since that's all set behind-the-scenes rather than by the user now
-  removing references in datasets to SVM

Additionally, this includes tangential changes now that the concept of multiple trainers is gone: 
- `AdvancedTrainingSettings` without the functionality to select a trainer was basically a dupe of `TrainingSettings` so it's been deleted 
-  `SetKValue` was extracted to its own component and included, but currently hidden in `TrainingSettings` 
- "selectTrainer" as a panel name was changed to "trainingSettings" since trainer selection is no longer supported


src/App.js
```diff
@@ -185,7 +185,7 @@ class App extends Component {
           </BodyContainer>
         )}
 
-        {currentPanel === "selectTrainer" && (
```
Erin007, line 188: This will impact levelbuilder mode parameters, yes? Should we hold off on this change or coordinate on the Code Studio side to get it in? I think it will mean updating level params already in the wild.

breville: I'm not sure that it will affect any of the params in the wild.  My understanding is that everything that currently exists in curriculum is now inventoried in https://github.com/code-dot-org/ml-playground/pull/138, so we could verify against that.

breville (approved):
Nice cleanup!

made-line (approved):
LGTM!

src/App.js
dancodedotorg: >My understanding is that everything that currently exists in curriculum is now inventoried in #138, so we could verify against that.

Hi. Eavesdropping. Yes - this is true - PR 138 now accurately reflects levelbuilder parameters. So if each of those new dropdown options looks good, then we should feel confident that this won't break anything on levelbuilder. And if it does, I can blame myself since I'm the one who made that PR =P

PR #141, opened 2021-04-13
Last images for datasets

moneppo:


Erin007 (approved):
teeny tiny nit: for consistency, can we make "auto-mpg" -> "auto_mpg" otherwise, looks great. Thanks for finding these!

PR #143, opened 2021-04-14
Remove unused code

breville:
Remove some unused code initially discovered by using Chrome's Coverage tab.


src/datasetManifest.js
```diff
@@ -250,9 +250,3 @@ export function getAvailableDatasets(specificDatasets) {
   // Sort dataset names by alphabetical order.
   return availableDatasets.sort((a, b) => a.name.localeCompare(b.name));
 }
-
-export function getDefaultLabelForDataset(datasetId) {
-  return allDatasets.find(dataset => {
-    return dataset.id === datasetId;
-  });
-}
```
Erin007, line 258: Hmmm -  there are levels in the curriculum where a default label is set (preference, for example, in the food datasets. How do we do that without this code?

src/redux.js
```diff
@@ -92,20 +90,12 @@ export function setSelectedTrainer(selectedTrainer) {
   return { type: SET_SELECTED_TRAINER, selectedTrainer };
 }
 
-export function setKValue(kValue) {
-  return { type: SET_K_VALUE, kValue };
-}
```
Erin007, line 97: I think we can also remove line 36 `const SET_K_VALUE = "SET_K_VALUE"; and then the kValue in initial state and also where we reference that state in the algorithm.  I'd even prefer if deleting `setKValue` and its tendrils lived in its own PR.

src/redux.js
```diff
@@ -92,20 +90,12 @@ export function setSelectedTrainer(selectedTrainer) {
...
 export const setColumnsByDataType = (column, dataType) => ({
   type: SET_COLUMNS_BY_DATA_TYPE,
   column,
   dataType
 });
 
-export function setSelectedFeatures(selectedFeatures) {
-  return { type: SET_SELECTED_FEATURES, selectedFeatures };
-}
-
```
Erin007, line 108: Similar to above, can you remove the action types references for each of these that you're deleting, please.

src/redux.js
```diff
@@ -443,12 +421,6 @@ export default function rootReducer(state = initialState, action) {
       trainedModel: action.trainedModel
     };
   }
-  if (action.type === SET_TRAINED_MODEL_DETAILS) {
-    return {
-      ...state,
-      trainedModelDetails: action.trainedModelDetails
-    };
-  }
   if (action.type === SET_TRAINED_MODEL_DETAIL) {
```
Erin007, line 424: Just clarifying, `setTrainedModelDetails` and `setSelectedFeatures` can both be removed because we edit the individual fields for those objects?

src/redux.js
```diff
@@ -696,15 +658,6 @@ export function getOptionFrequencies(state, column) {
...
 
-export function getOptionFrequenciesByColumn(state) {
-  let optionFrequenciesByColumn = {};
-  getSelectedCategoricalColumns(state).map(
-    column =>
-      (optionFrequenciesByColumn[column] = getOptionFrequencies(state, column))
-  );
-  return optionFrequenciesByColumn;
-}
-
```
Erin007, line 707: This can be deleted because we now show the bar charts instead of these numbers?

src/datasetManifest.js
breville: Looks like we set it here: https://github.com/code-dot-org/ml-playground/blob/main/src/jsonReaderWrapper.js#L38

src/redux.js
breville: It looks like we do still have actions that modify them, at https://github.com/code-dot-org/ml-playground/blob/0c27d7ee37550eda2b7fc11b7821cf217507bd1e/src/redux.js#L452 and https://github.com/code-dot-org/ml-playground/blob/0c27d7ee37550eda2b7fc11b7821cf217507bd1e/src/redux.js#L339-L356 respectively.

src/redux.js
breville: Good catch, will do.

src/redux.js
breville: I think so.  I discovered this uncalled code through Chrome's coverage feature, and then couldn't find any remaining calls to this function.

src/redux.js
breville: Prepped a standalone PR in https://github.com/code-dot-org/ml-playground/pull/146 which can be merged first.

src/redux.js
breville: It turns out that `getOptionFrequencies`, immediately above, takes a `column`, so it seems to serve the right purpose for `ColumnInspector`.

Erin007 (approved):
Awe - one of the other tested functions and it's time to say goodbye.  Thanks for the clean up!

PR #144, opened 2021-04-15
Add reorganized features and label to dataToSave

Erin007:
Based on the [plan](https://docs.google.com/document/d/1HSoypNk8O-8hSJQ0eZfgkOoDtT9703NNFlQXoQK1UxY/edit#bookmark=id.xs7n38y6s9vg) for re-organizing `dataToSave` for long-term maintainability and ease of use in App Lab, I added two new fields: `label` and `features` to metadata. 

Screen Shot 2021-04-14 at 8 31 06 PM

This step is not risky because it's purely additive. Up next is using these values in App Lab. I checked that it saves data as expected for numerical and categorical features and labels.

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -751,7 +751,7 @@ export function getSelectedColumnDescriptions(state) {
 }
 
 export function getColumnDescription(state, column) {
-  if (!state.metadata || !state.metadata.fields) {
+  if (!state.metadata || !state.metadata.fields || !column) {
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


breville (approved):

made-line:
Just for my own understanding, doesn't the existing `dataToSave.selectedFeatures = state.selectedFeatures;` do the same thing as `dataToSave.features = getFeaturesToSave(state);`?

Erin007:
> Just for my own understanding, doesn't the existing `dataToSave.selectedFeatures = state.selectedFeatures;` do the same thing as `dataToSave.features = getFeaturesToSave(state);`?

Good question! They are similar, but not the same. 

`dataToSave.selectedFeatures` is an array of feature names, something like `["feathers", "legs"]`. 

`dataToSave.features` is an array of feature objects, something like ```[{id: "feathers", description: "Whether it has feathers", values: ["yes", "no"]}, {id: "legs", description: "How many legs it has", max: 8, min: 0}]```

The goal is to ultimately make it easier to iterate through the selected features once they're in App Lab and not have to do a series of look-ups to the `featureNumberKey`, `columns`, and `extremums` data fields for each feature.

made-line:
Ah, thank you for the clarification!

PR #145, opened 2021-04-15
Style data card

breville:
This updates the styling of the data card.

It also tidies up the format of the JSON, removes unused fields from the JSON, reduces the size of the upload UI, and doesn't reload data when clicking the same item more than once.

Screen Shot 2021-04-15 at 11 31 52 AM


public/datasets/auto_mpg.json
```diff
@@ -1 +1,51 @@
...
+{
+  "card": {
+    "description": "A small dataset of different car statistics, originally from 1983, used to predict their MPG (miles per gallon). This dataset is unlikely to get great accuracy scores on a KNN regression, but it could make a useful model nonetheless: how can you take a general pattern in the model and use that to determine which features contribute to higher miles per gallon? How might you change the dataset to get a more accurate model - for example, changing MPG from a numerical column to a categorical one with intervals of 5 MPG?",
```
Erin007, line 3: This description is interesting, because I the others don't talk about accuracy or pose questions like this. For example, it feels weird to ask how the dataset might be changed when we aren't giving students the opportunity to really change the datasets. I wonder if it needs to be more neutral?

public/datasets/auto_mpg.json
```diff
@@ -1 +1,51 @@
...
+      "type": "numerical",
+      "id": "horsepower",
+      "description": "The power of the car."
+    },
+    {
+      "type": "numerical",
+      "id": "weight",
+      "description": "The car's weight in pounds."
+    },
+    { "type": "numerical", "id": "acceleration", "description": "" },
```
Erin007, line 39: No description, here?

public/datasets/foods.json
```diff
@@ -1,50 +1,50 @@
...
+      "type": "numerical"
+    },
+    {
+      "id": "sports",
+      "type": "categorical"
+    },
+    {
+      "id": "type_sports",
+      "type": "categorical"
+    },
```
Erin007, line 47: No descriptions for these?

public/datasets/global_happiness_raw.json
```diff
@@ -1,4 +1,15 @@
 {
+  "card": {
+    "description": "2019 Global Happiness report rankings, where countries are rated on different attributes that contribute to a total happiness score then ranked. The features in this dataset are taken from the years 2008 to 2018, which were averaged together for the report to create scores. This 'raw' data contains more rows than the 'Global Happiness 2019' dataset, but may give different results.",
```
Erin007, line 3: Do we want to keep in the mention of the other 'Global Happiness' dataset if it's no longer in the tool? Wouldn't want to send students on a goose chase.

public/datasets/heart.json
```diff
@@ -1 +1,63 @@
...
+    "source": "UCI (http://archive.ics.uci.edu/ml/datasets/Heart+Disease)",
+    "lastUpdated": "1988",
+    "context": {
+      "createdBy": "1. Hungarian Institute of Cardiology. Budapest: Andras Janosi, M.D., University Hospital, Zurich, Switzerland: William Steinbrunn, M.D., University Hospital, Basel, Switzerland: Matthias Pfisterer, M.D., V.A. Medical Center, Long Beach and Cleveland Clinic Foundation: Robert Detrano, M.D., Ph.D.",
+      "potentialUses": "Diagnose whether someone may be at risk for heart disease based on various factors.",
+      "potentialMisuses": ""
+    }
+  },
+  "fields": [
+    { "type": "numerical", "id": "\ufeffage", "description": "" },
```
Erin007, line 13: What's this field? Does it need the \ and should it have a description?

Erin007 (approved):
Code looks good to me and the breathing room in the card makes it much easier to scan!

 I had a few questions about the content in JSON dataset files - mostly why some columns don't have descriptions and if we want to revise some of the dataset descriptions. I don't feel super strongly about the comments, so I'll let you decide whether to pursue here, as follow up or if they might be better questions for @moneppo?

made-line (approved):
Just to clarify, did Mike want the source URLs removed? https://trello.com/c/Sw0hjqj4/148-dataset-info-is-confusing

breville:
@Erin007 Great catches.  I'm content agnostic but will open an item to track these improvements to our metadata.

@made-line I believe we just don't need them to be clickable.

PR #150, opened 2021-04-20
ort selected features

Erin007:
Sort selected features.  Order of feature selection, was curiously sometimes affecting accuracy. Investigation documentation [here](https://docs.google.com/document/d/1Qi8A4JV5ymB74o2pfuAUo5fZpKp2Iu_FxZ-nQCUX0Og/edit#).

Order of feature selection should not affect the outcome of the algorithm, and it's confusing to teachers and students when it does.  Consistency between runs with the same selected features is important. To preserve that consistency we're sorting the selected features at the recommendation of the AI consultant, which unblocks the curriculum team for this specific example: 

BEFORE: 

https://user-images.githubusercontent.com/12300669/115455339-d3f16380-a1ef-11eb-98e9-5f3d1febc792.mp4

AFTER: 

https://user-images.githubusercontent.com/12300669/115455181-9c82b700-a1ef-11eb-9ba8-4f90b5139fd8.mov

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -297,17 +297,16 @@ export default function rootReducer(state = initialState, action) {
       }
     };
   }
-
   if (action.type === ADD_SELECTED_FEATURE) {
     if (!state.selectedFeatures.includes(action.selectedFeature)) {
+      const newFeatures = [...state.selectedFeatures, action.selectedFeature];
       return {
         ...state,
-        selectedFeatures: [...state.selectedFeatures, action.selectedFeature],
+        selectedFeatures: newFeatures.sort(),
         currentColumn: undefined
       };
     }
   }
-
   if (action.type === REMOVE_SELECTED_FEATURE) {
     return {
       ...state,


Erin007:
This stopgap solution to ensure consistency between runs masks a deeper underlying problem that still needs to be root caused and addressed. In this case, the column with the high predictive power (cherry) also happens to be sorted as the first feature alphabetically, and the "faulty" column (whipped cream) is second. In other datasets in which the "faulty" column  is alphabetically first (e.g. cheese in the taco dataset) sorting would ensure the unexpected _lower_ accuracy outcome. Without sorting Cheese then Chicken -> 64%, Chicken then Cheese -> 94%; with sorting Chicken + Cheese regardless of order would be consistently 64%, not what we want. 

@dancodedotorg I'm going to focus on digging in to see if I figure out what's going on to cause the inconsistency, but wonder how you feel in the meantime: merge or no merge? What's easier for you to message to pilot teachers?

dancodedotorg:
@Erin007 - thank you for checking!
> with sorting Chicken + Cheese regardless of order would be consistently 64%, not what we want.

^^ This is a strong pass from me, since it directly impacts learning goals. We want students to think "strong relationship --> stong model" - if introducing this fix muddles that intuition, then we shouldn't merge.

In terms of messaging to teachers: I meet with them tonight (4/20), so it's easy for me to let them know that this is a known bug and they can have this in their back pocket when working with students. I'd rather do this - let the bug live in the wild for now, and tell teachers to be aware of it - than try and implement the sorting fix that could lead to inaccurate models.

Erin007:
> @Erin007 - thank you for checking!
> 
> > with sorting Chicken + Cheese regardless of order would be consistently 64%, not what we want.
> 
> ^^ This is a strong pass from me, since it directly impacts learning goals. We want students to think "strong relationship --> stong model" - if introducing this fix muddles that intuition, then we shouldn't merge.
> 
> In terms of messaging to teachers: I meet with them tonight (4/20), so it's easy for me to let them know that this is a known bug and they can have this in their back pocket when working with students. I'd rather do this - let the bug live in the wild for now, and tell teachers to be aware of it - than try and implement the sorting fix that could lead to inaccurate models.

Sounds good. Thanks! Once I thought through implications, I was strongly doubting this path too. I'll close this PR and keep digging for the real issue.

PR #151, opened 2021-04-20
Summary screen

made-line:
This PR includes the following changes:
* Decoupled the save and finish functions. 
* Added a Summary screen after the Save Model screen which includes:
  * the model card
  * a loading spinner/message to indicate the model was saved.
* Finish button is enabled when model is saved. 
 

https://user-images.githubusercontent.com/40412372/115465922-2a58a500-a1e4-11eb-96b8-82528c5be665.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-svg-core": "^1.2.25",
     "@fortawesome/free-regular-svg-icons": "^5.11.2",
-    "@fortawesome/free-solid-svg-icons": "^5.11.2",
+    "@fortawesome/free-solid-svg-icons": "^5.15.3",
     "@fortawesome/react-fontawesome": "^0.1.7",
     "chart.js": "^2.9.4",
     "ml-svm": "^2.1.2",
--- a/src/App.js
+++ b/src/App.js
@@ -11,6 +11,7 @@ import Results from "./UIComponents/Results";
 import Predict from "./UIComponents/Predict";
 import SaveModel from "./UIComponents/SaveModel";
 import ModelSummary from "./UIComponents/ModelSummary";
+import ModelCard from "./UIComponents/ModelCard";
 import { styles } from "./constants";
 import { connect } from "react-redux";
 import {
@@ -19,7 +20,6 @@ import {
   validationMessages,
   getTrainedModelDataToSave
 } from "./redux";
-import Statement from "./UIComponents/Statement";
 
 class PanelButtons extends Component {
   static propTypes = {
@@ -230,14 +230,11 @@ class App extends Component {
 
         {currentPanel === "modelSummary" && (
           <BodyContainer>
-            {/* <ContainerFullWidth>
-              <ModelSummary />
-            </ContainerFullWidth> */}
             <ContainerLeft>
-              <Statement />
+              <ModelSummary />
             </ContainerLeft>
             <ContainerRight>
-              <ModelSummary />
+              <ModelCard />
             </ContainerRight>
           </BodyContainer>
         )}
--- a/src/App.js
+++ b/src/App.js
@@ -19,6 +19,7 @@ import {
   validationMessages,
   getTrainedModelDataToSave
 } from "./redux";
+import Statement from "./UIComponents/Statement";
 
 class PanelButtons extends Component {
   static propTypes = {
@@ -229,9 +230,15 @@ class App extends Component {
 
         {currentPanel === "modelSummary" && (
           <BodyContainer>
-            <ContainerFullWidth>
+            {/* <ContainerFullWidth>
               <ModelSummary />
-            </ContainerFullWidth>
+            </ContainerFullWidth> */}
+            <ContainerLeft>
+              <Statement />
+            </ContainerLeft>
+            <ContainerRight>
+              <ModelSummary />
+            </ContainerRight>
           </BodyContainer>
         )}
 
--- a/src/App.js
+++ b/src/App.js
@@ -38,7 +38,10 @@ class PanelButtons extends Component {
     if (this.props.panelButtons.next.panel === "modelSummary") {
       this.props.startSaveTrainedModel(this.props.dataToSave);
       this.props.setCurrentPanel(this.props.panelButtons.next.panel);
-    } else if (this.props.panelButtons.next.panel === "finish") {
+    } else if (
+      this.props.panelButtons.next.panel === "continue" ||
+      this.props.panelButtons.next.panel === "finish"
+    ) {
       this.props.onContinue();
     } else {
       this.props.setCurrentPanel(this.props.panelButtons.next.panel);
--- a/src/App.js
+++ b/src/App.js
@@ -10,6 +10,7 @@ import TrainModel from "./UIComponents/TrainModel";
 import Results from "./UIComponents/Results";
 import Predict from "./UIComponents/Predict";
 import SaveModel from "./UIComponents/SaveModel";
+import ModelSummary from "./UIComponents/ModelSummary";
 import { styles } from "./constants";
 import { connect } from "react-redux";
 import {
@@ -34,9 +35,10 @@ class PanelButtons extends Component {
   };
 
   onClickNext = () => {
-    if (this.props.panelButtons.next.panel === "save") {
+    if (this.props.panelButtons.next.panel === "modelSummary") {
       this.props.startSaveTrainedModel(this.props.dataToSave);
-    } else if (this.props.panelButtons.next.panel === "continue") {
+      this.props.setCurrentPanel(this.props.panelButtons.next.panel);
+    } else if (this.props.panelButtons.next.panel === "finish") {
       this.props.onContinue();
     } else {
       this.props.setCurrentPanel(this.props.panelButtons.next.panel);
@@ -218,6 +220,14 @@ class App extends Component {
           <BodyContainer>
             <ContainerFullWidth>
               <SaveModel />
+            </ContainerFullWidth>
+          </BodyContainer>
+        )}
+
+        {currentPanel === "modelSummary" && (
+          <BodyContainer>
+            <ContainerFullWidth>
+              <ModelSummary />
             </ContainerFullWidth>
           </BodyContainer>
         )}
--- a/src/App.js
+++ b/src/App.js
@@ -36,17 +36,19 @@ class PanelButtons extends Component {
   };
 
   onClickNext = () => {
-    if (this.props.panelButtons.next.panel === "modelSummary") {
-      this.props.startSaveTrainedModel(this.props.dataToSave);
-      this.props.setCurrentPanel(this.props.panelButtons.next.panel);
-    } else if (
+    if (
       this.props.panelButtons.next.panel === "continue" ||
       this.props.panelButtons.next.panel === "finish"
     ) {
       this.props.onContinue();
-    } else {
-      this.props.setCurrentPanel(this.props.panelButtons.next.panel);
+      return;
     }
+
+    if (this.props.panelButtons.next.panel === "modelSummary") {
+      this.props.startSaveTrainedModel(this.props.dataToSave);
+    }
+
+    this.props.setCurrentPanel(this.props.panelButtons.next.panel);
   };
 
   render() {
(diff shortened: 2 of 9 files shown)


breville:
Just aesthetically, I think the text should be vertically centered in the left panel, rather than being stuck at the very top.

I think it would also be good to ensure that the spinner shows for a minimum amount of time (e.g. 1 or 2 seconds) to ensure that all users get to see it.  Otherwise, users on fast connection will see a flicker of it, but won't have enough time to understand what they're seeing.

package.json
```diff
@@ -73,7 +73,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-svg-core": "^1.2.25",
     "@fortawesome/free-regular-svg-icons": "^5.11.2",
-    "@fortawesome/free-solid-svg-icons": "^5.11.2",
+    "@fortawesome/free-solid-svg-icons": "^5.15.3",
```
breville, line 76: Out of curiosity, was this bump needed for the spinner?

src/App.js
```diff
@@ -34,13 +36,19 @@ class PanelButtons extends Component {
   };
 
   onClickNext = () => {
-    if (this.props.panelButtons.next.panel === "save") {
-      this.props.startSaveTrainedModel(this.props.dataToSave);
-    } else if (this.props.panelButtons.next.panel === "continue") {
+    if (
+      this.props.panelButtons.next.panel === "continue" ||
+      this.props.panelButtons.next.panel === "finish"
```
breville, line 41: Could also use `includes` here...

src/redux.js
```diff
@@ -1191,8 +1198,8 @@ export function getPanelButtons(state) {
     next = isPanelAvailable(state, "dataDisplayLabel")
       ? { panel: "dataDisplayLabel", text: "Continue" }
       : isPanelAvailable(state, "dataDisplayFeatures")
-      ? { panel: "dataDisplayFeatures", text: "Continue" }
-      : null;
+        ? { panel: "dataDisplayFeatures", text: "Continue" }
+        : null;
```
breville, line 1202: Why did this indenting change?

src/redux.js
```diff
@@ -1228,8 +1235,13 @@ export function getPanelButtons(state) {
...
-    next = isPanelAvailable(state, "save")
-      ? { panel: "save", text: "Finish" }
+    next = isPanelAvailable(state, "modelSummary")
+      ? { panel: "modelSummary", text: "Save" }
+      : null;
+  } else if (state.currentPanel === "modelSummary") {
+    prev = { panel: "saveModel", text: "Back" };
+    next = isPanelAvailable(state, "finish")
+      ? { panel: "finish", text: "Finish" }
       : null;
```
breville, line 1245: What's the user experience if the save fails?

src/UIComponents/ModelSummary.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+    this.state = {
+      isLoading: true
+    };
+  }
+
+  componentDidMount() {
+    // Add a timer to simulate loading when saving a model.
+    let loadSpinner = () => this.setState({ isLoading: false });
+    setTimeout(loadSpinner, 1500);
```
breville, line 25: What happens if the save takes longer than 1.5 seconds?  Do we keep showing the spinner until it's actually done?

Erin007:
Ah, ok, I see what you mean about this page feeling empty.  I don't think we need to reserve the entire left side of the screen for the "Your model has been saved" message - the layout in the dialog in App Lab is different because of the dropdown, the additional buttons and the context. I think for this summary screen, couldn't we just have the model card take up the majority of the screen and be centered? 

Screen Shot 2021-04-21 at 11 46 30 AM

dancodedotorg:
👋 If there's a "bunch of empty space" problem, I'd encourage us to think about how to include A.I. Bot on the screen. One overwhelming piece of positive feedback from our classroom visits were **people like A.I. Bot**, and this feels like an easy win to include it back in the mix.

Also, A.I. Bot is the representation of your model, and the model card is the documentation for your model, so it makes sense to me to have both of them on this same screen.

I know there was a version with A.I. bot and the typewriter text - maybe that's worth revisiting? Or finding another way to include A.I. Bot on this screen.

Erin007:
Will we want to set a default for dynamic instructions that goes with this new screen? Something like "You've successfully trained and saved your move. Review your model details and click Finish to use your trained model in an app."

src/App.js
made-line: I refactored it with `includes`. Thanks!

src/redux.js
made-line: I formatted the file with Prettier. I changed it back.

package.json
made-line: It looks like it wasn't needed. I reverted it back and the spinner still shows.

src/UIComponents/ModelSummary.jsx
made-line: I removed the timer functionality. It shows the spinner until the model is saved now.

src/redux.js
made-line: Good call! I changed this functionality. If the save fails, they'll see a failure message and can click Save to try again. If it succeeds, the Save button becomes Continue.

made-line:
Thank you all for your feedback! I made a lot of changes and added dynamic instructions in https://github.com/code-dot-org/code-dot-org/pull/40303.

src/redux.js
breville: These buttons are designed as navigation buttons rather than action buttons with no navigation.  I think that once Save is pressed, a successful save should navigate to the next screen rather than turn into a different button.

breville:
- Have you been able to test the code path when we get a server error?  I hit upon one and saw this in the debug console; not sure if it was expected:
Screen Shot 2021-04-29 at 9 47 51 AM

- The spinner & message seem to be too small and overlap existing UI in some cases:
Screen Shot 2021-04-29 at 9 38 33 AM
There are a few ways we could address this, including a modal popup, or leaving some space for them in the left panel.  But one thing we could try is to put them in between the prev & next nav buttons, since we have quite a bit of room there.

- As noted in a separate comment, I think we should have the prev & next nav buttons always do a navigation.  I'd suggest that after pressing "Save", both buttons are disabled while we show the spinner, and then if the save succeeds, we navigate to the final screen, or if it fails, we reactivate the buttons as appropriate.

- Should the final screen's next button now say "Finish"?  Given that we're just going to the next level, I wonder if "Continue" is the better choice now.

- It's good to see the bot again.  But we should have a background of some kind, or use the bot image with the outlines, since currently its edges vanish against the white background.

- Is the bot best-placed in the bottom-left corner?  It feels like it's hiding a bit.  One thing we could consider is having the bot show on the model card over in applab so that this screen feels more connected to that screen; we can work out how to be consistent between the two.
 
- Can we show the Statement on the final screen?  It feels like it would be good to remind the student what they trained.

- The abalone data set that I happened to pick has a ton of "possible values".  Do we want to show such a long piece of text?
Screen Shot 2021-04-29 at 9 54 56 AM

Happy to talk through these notes, and also to help out on the PR if desired, but thought I'd collect here first.

src/App.js
```diff
@@ -26,25 +29,38 @@ class PanelButtons extends Component {
...
-      this.props.setCurrentPanel(this.props.panelButtons.next.panel);
+      return;
+    }
+
+    if (this.props.currentPanel === "saveModel") {
+      this.props.startSaveTrainedModel(this.props.dataToSave);
+      return;
     }
+
+    this.props.setCurrentPanel(this.props.panelButtons.next.panel);
```
breville, line 51: nit: we could use `else if` and `else` to avoid the `return` statements.

src/App.js
```diff
@@ -26,25 +29,38 @@ class PanelButtons extends Component {
...
   };
 
   render() {
-    const { panelButtons } = this.props;
+    const { panelButtons, saveStatus } = this.props;
+
+    let loadStatus = !["success", "failure"].includes(saveStatus)
+    ? ( <FontAwesomeIcon icon={faSpinner} />)
+    : (saveMessages[saveStatus]);
```
breville, line 59: nit: our current formatting would indent these lines.

src/redux.js
```diff
@@ -1110,6 +1118,16 @@ function isPanelAvailable(state, panelId) {
...
   }
 
+  if (panelId === "modelSummary") {
+    if ([undefined, ""].includes(state.trainedModelDetails.name)) {
+      return false;
+    }
+
+    // if (state.saveStatus === "started") {
+    //   return false;
+    // }
```
breville, line 1128: nit: remove?

src/UIComponents/ModelCard.jsx
```diff
@@ -0,0 +1,156 @@
+/* React component to handle displaying the model card. */
+import PropTypes from "prop-types";
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import { styles } from "../constants";
+import { getSummaryStat,
```
breville, line 6: nit: our current styling would also have this first entry on a new line.  We should be consistent.

src/constants.js
```diff
@@ -508,5 +508,65 @@ export const styles = {
...
   saveInputsWidth: {
     width: "95%"
+  },
+
+  modelCardContainer: {
+    backgroundColor: "rgb(198, 202, 205)",
+    borderRadius: 5,
+    padding: 20,
+    whiteSpace: "normal",
+    overflowY: "scroll",
```
breville, line 518: We generally switched to `"auto"` so that Windows machines in particular only show a scrollbar when the content is actually scrollable.

src/App.js
```diff
@@ -26,25 +29,38 @@ class PanelButtons extends Component {
...
     }
+
+    this.props.setCurrentPanel(this.props.panelButtons.next.panel);
   };
 
   render() {
-    const { panelButtons } = this.props;
+    const { panelButtons, saveStatus } = this.props;
+
+    let loadStatus = !["success", "failure"].includes(saveStatus)
```
breville, line 57: Does this imply `saveStatus` is `"started"` or is there another possibility?

src/UIComponents/Predict.jsx
```diff
@@ -46,12 +46,14 @@ class Predict extends Component {
           <div style={styles.scrollingContents}>
             <form>
               {this.props.selectedNumericalFeatures.map((feature, index) => {
+                // toFixed converts a number to a string and rounds to two decimal places.
```
Erin007, line 49: Not sure we need these comments and the one below?

Erin007:
Screen Shot 2021-05-04 at 10 30 31 AM
AI Bot doesn't show for me locally - is there a trick to fix the image?

Erin007:
I don't think the save failure message is showing when it should. In the case where the metadata field is too long (movie dataset predicting budget based on director and star) the save hits a 500 error and I'd expect the message "Your model did not save" to appear but instead the spinner just disappears, the save button reappears and there's no error message. 

https://user-images.githubusercontent.com/12300669/117024717-3f473500-acc8-11eb-9e10-cc2697f8aa7d.mov


Screen Shot 2021-05-04 at 11 00 43 AM

Separately, we can better handle this case in the database and further customizing what the API returns, but wondered if the Save Model screen should've behaved differently in this scenario.

src/redux.js
made-line: I missed this. Thanks!

src/constants.js
made-line: Fixed. Thanks!

src/App.js
made-line: `saveStatus` by default is `undefined`. This covers the cases where the save hasn't finished and it'd still work if we wanted to add more in between states in the future. :crossed_fingers: What do you think?

breville: I think we could refactor this section of code a bit to move some of the specific knowledge about possible values of `saveStatus`, and what to do with it, out of the UI and into a Redux helper function or two.  Then we simply have a prop (or two) which tells us whether we show a spinner, whether we show a message (and what it is), and then we avoid rendering `{undefined}` for nothing as an added bonus.  This also means that if we add another state to `saveStatus` later, we don't need to hardcore references to it in our views, and rather centralise that management in `redux.js`.

made-line: Good call! I refactored it. Let me know if this update is on track with what you were thinking.

breville (approved):

src/constants.js
```diff
@@ -25,7 +25,7 @@ export const ModelNameMaxLength = 150;
 
 export const saveMessages = {
   success: "Your model was saved!",
-  failure: "There was an error. Your model did not save.",
+  failure: "There was an error. Your model did not save. Please try again.",
   name: "Please name your model."
```
Erin007, line 29: I think we can remove name if it's never shown

PR #155, opened 2021-04-22
Update strings on Save Model screen

made-line:
[Trello task](https://trello.com/c/fjRoAAzM/175-update-text-under-how-can-the-model-be-misused)

Before: 
Screen Shot 2021-04-21 at 6 37 25 PM

After: 
Screen Shot 2021-04-23 at 5 27 19 PM


src/UIComponents/SaveModel.jsx
```diff
@@ -51,19 +50,31 @@ class SaveModel extends Component {
...
+      description: (
+        <div>
+          Describe any situations where this model could potentially be misused,
+          or any places where bias could potentially show up in the model.
+          Important questions to consider are:
+          <ul style={{ marginBottom: 0 }}>
+            <li>Is there enough data to create an accurate model?</li>
+            <li>Does the data represent all possible users and scenarios?</li>
+          </ul>
+        </div>
```
breville, line 76: I know we're not localising just yet, but this will be particularly difficult since I don't think we want to put JSX through the pipeline.  I wonder if it's worth already breaking this up into a few separate strings with no formatting involved...

Erin007: Seconded. I think we also want to avoid inline styling, per https://github.com/code-dot-org/code-dot-org/blob/staging/STYLEGUIDE.md#react

dancodedotorg:
Love these updates! Love that the headers are now bolded to make it easier to identify the parts of the model card, and love the additional sentences to help steer students in the right direction for how to answer these prompts.

(I don't have any useful engineering feedback... but design-wise, looks awesome)

src/UIComponents/SaveModel.jsx
made-line: Thank you both! I reviewed the style guide and implemented this feedback.

src/UIComponents/SaveModel.jsx
```diff
@@ -157,8 +163,14 @@ class SaveModel extends Component {
               {this.getUsesFields().map(field => {
                 return (
                   <div key={field.id} style={styles.cardRow}>
-                    <label>{field.text}</label>
+                    <label style={styles.bold}>{field.text}</label>
                     <div>{field.description}</div>
+                    {field.misuseQuestionOne && (
```
Erin007, line 168: One nit but not blocking: could these field names be more generic since they're part of a generic component? e.g. `descriptionDetailOne`. It'd make it slightly more extensible, let's say if in the future we also wanted to add something similar to the Intended Use description.

made-line: Will keep this in mind going forward. I updated these variable names. Thank you!

PR #156, opened 2021-04-22
Fix Results' stick headers

breville:
The initial work to make the Results header sticky worked for categorical columns, but made an assumption that the first row's height was going to be 30px when setting the top sticky position of the second row.  However, when there is a numerical column, the first row is actually taller than that.

This is a simple fix that always assumes a taller first row, which happens with a numerical column.  Now, the second header row remains in a fixed location rather than scrolls a bit.

A more comprehensive fix would be to adjust the height depending whether that taller first row is needed, but we can consider that as we further refine this screen.

With numerical column:

Screen Shot 2021-04-22 at 12 19 52 PM

Without numerical column:

Screen Shot 2021-04-22 at 12 24 59 PM

### DIFF
--- a/src/UIComponents/ResultsTable.jsx
+++ b/src/UIComponents/ResultsTable.jsx
@@ -49,7 +49,7 @@ class ResultsTable extends Component {
                 >
                   <span style={styles.largeText}>{"Actual"}</span>
                   {this.props.isRegression && (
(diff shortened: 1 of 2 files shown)


Erin007 (approved):
Thank you!

PR #157, opened 2021-04-22
Updated DataDisplay UI

breville:
This is a significant update to the `DataDisplay` UI in which the student explores columns, chooses a label, and chooses features.

The impetus was going through the curriculum and realising that there is a big emphasis placed on examining the graph for a column and then using that to decide whether the column should be chosen as a label or feature.  The current UI with the `PhraseBuilder` in the right-hand panel was quite nice, but the student experience was looking at graphs in popups, and then navigating over to find the same column name in a dropdown in the `PhraseBuilder`, often having to close the popup first to get access to that panel.

A more natural flow would be to examine a graph and then immediately make the choice to choose that column as a label or feature.  We could have kept the popup, but without the `PhaseBuilder`, the right-hand panel became available again, and it avoids a lot of the subtle issues relating to screen management with the popup (such as positioning it correctly, not causing the screen to scroll if it's particularly long, etc.)

With that in mind, I moved the `ColumnInspector` out of the popup and back into the right-hand panel.  I then brought back a very early implementation we had, in which there is a button to choose the current column.  That early idea also had a button to un-choose the column, but it seemed more intuitive to add a close X icon to the `Statement` at the top instead, allowing the student to actually interact with the `Statement`, rather than just have it as a reflection of something done elsewhere.

It got slightly confusing having light blue as the highlight colour when choosing a label, so I added light red variants of the highlighted- and selected-column colour schemes.

https://user-images.githubusercontent.com/2205926/115659033-ca025a00-a2ee-11eb-801d-4c3d073ae136.mov


markabarrett (approved):
Compared to what we had in the previous version, this is a really nice evolution and improvement to how we handle selecting labels and features in AI Lab!

src/UIComponents/DataDisplay.jsx
```diff
@@ -4,18 +4,15 @@ import React, { Component } from "react";
...
 class DataDisplay extends Component {
   static propTypes = {
-    data: PropTypes.array,
-    setCurrentColumn: PropTypes.func,
-    setColumnRef: PropTypes.func
+    data: PropTypes.array
   };
 
   render() {
-    const { data, setCurrentColumn, setColumnRef } = this.props;
```
Erin007, line 18: I like that this simplifies both the UI and the code by no longer having to deal with the refs

src/constants.js
```diff
@@ -370,19 +375,58 @@ export const styles = {
...
+    display: "inline-block",
+    position: "relative",
+    lineHeight: 1.3
+  },
+
+  statementDeleteIcon: {
+    position: "absolute",
+    top: 0,
+    right: 0,
+    transform: 'translateX(50%) translateY(-50%)',
```
Erin007, line 410: ah, I see. clever construction of this composite icon

Erin007:
Code changes look good.  Nothing blocking. I really like this direction and feels much cleaner than all the pop up in my opinion.  

A few thoughts while playing with this new layout: 

- On the label selection screen, what do you think about having just the first part of the Mad Lib? "Predict..." then on the following screen "based on..." shows up. It felt a little funny to me to fill in one ellipsis but still have another empty one when I moved on. 
- Graphs on the side rather than overlays are way more pleasant to use. The animation when switching columns is 💯  and it makes is easier to look at the content of a column and the graph. 
- I wonder if we should bump up the graph limit - for example there are 7 animal classes in the zoo dataset and it would be nice to see graphs for them. Similarly if we don't show a graph because there are too many possibilities should we at least list what the potential values in that column are (at least up to a certain #, say 20?)
- What are we doing with the pre-training screen? This is a bigger conversation, but relates to this change because if we do end up keeping it, we could think about keeping the Mad Lib consistent on that panel as well.  Right now the content is there but worded and styled differently. 
Screen Shot 2021-04-29 at 12 32 25 PM

Erin007 (approved):

src/UIComponents/Statement.jsx
```diff
@@ -25,20 +38,48 @@ class Statement extends Component {
...
+                      </div>
+                    </div>
+                  )}
+                </div>
+                {index < selectedFeatures.length - 1 && ", "}
+              </span>
+            );
+          })}
+          {selectedFeatures.length === 0 && (
+            <span style={styles.statementFeature}>...</span>
```
made-line, line 80: Very cool! Just wondering, how come you replaced lines 35-37 with this?

dancodedotorg (approved):

made-line (approved):
Looks great!

src/UIComponents/Statement.jsx
breville: Previously we wrapped the whole set of features, separated by commas, in `styles.statementFeature`.  Now, we wrap each individual feature (if there are any), in that style.

breville:
Thanks for the reviews.  I'm going to merge this initial work, but will go through all the open comments for potential followups.

PR #159, opened 2021-04-23
Rollup UI

breville:
This adds a rollup to the `Results` screen, and puts `ResultsDetails` in a popup.  The rollup shows the current and prior models for this invocation of AI Lab, with the accuracy of each.  The details of the most recent model are available in a popup.  There is support for an optional `requireAccuracy` parameter which specifies the percentage accuracy which the most recent run must reach before the user can continue beyond this screen; otherwise, they can navigate back to try again, creating a kind of game in which the user must find a combination of label and features that reaches the desired accuracy.

The `Results` rollup:

Screen Shot 2021-04-27 at 3 54 59 PM

The `ResultsDetails` popup:

Screen Shot 2021-04-27 at 3 55 06 PM

This video illustrates three iterations through training, with the third reaching the desired `requireAccuracy` of `50` (percent), at which point the user may progress to `SaveModel`.

https://user-images.githubusercontent.com/2205926/116192083-33fd7380-a6e2-11eb-9a82-908941c6aaee.mov


markabarrett (commented):
We might be able to make some improvements to the layout and presentation but this seems like a good way to motivate users to more deeply explore their feature choices and improve the accuracy of their model

src/constants.js
```diff
@@ -362,6 +351,36 @@ export const styles = {
...
+  },
+
+  resultsDetailsButton: {
+    fontSize: 14,
+    padding: "4px 12px",
+    margin: 0,
+    border: "none",
+    cursor: "pointer",
+    borderRadius: 5,
+    backgroundColor: "#61d2eb",
```
moneppo, line 385: #0094CA is the standard "blue" for buttons

src/constants.js
```diff
@@ -379,6 +433,11 @@ export const styles = {
     lineHeight: '50px'
   },
 
+  statementSmall: {
+    fontSize: 18,
+    paddingBottom: 6
+  },
+
   statementLabel: {
```
moneppo, line 441: I know this is a small thing but can we round the corners on each word and give a little more left/right padding?

src/constants.js
```diff
@@ -121,6 +121,28 @@ export const styles = {
...
+  panelPopup: {
+    padding: 10,
+    backgroundColor: "white",
+    overflow: "hidden",
+    height: "100%",
+    boxSizing: "border-box",
+    display: "flex",
+    flexDirection: "column",
+    //opacity: 0.7,
+    border: "solid 1px black",
```
moneppo, line 142: How difficult would it be to style this like our master modal style (https://www.figma.com/file/hVEnbfL3Ii13Nzsp0xdGfV/Code-Design-System-1.0?node-id=0%3A1800)?

breville: Have we implemented this anywhere yet, so that I can borrow the styling?

src/constants.js
breville: Is there any concern that it makes these look like buttons?

Screen Shot 2021-04-29 at 11 13 24 AM

src/constants.js
breville: Oh, fwiw, I tried picking a colour that would look different to our feature words on the left, thinking that this might be confusing:

Screen Shot 2021-04-29 at 11 22 17 AM

src/constants.js
Erin007: +1 to the concern that rounded corners make them look like buttons, also the square corners more closely match the column headings and anything we can do to reinforce that connection (e.g. like we did with color) feels like a good direction to head

made-line: I like the rounded corners. It matches the rounded corners of the model card. I think it's okay for the features to look like buttons in the mad lib since you can click on them to remove them.

Erin007 (approved):
I have a slight preference for extracting the accuracy threshold param from this PR since conceptually it feels like a different change to results roll up, but otherwise LGTM!

dancodedotorg (approved):

made-line:
Looks great! 

* Why are the details only available for the current model? Are we thinking it wouldn't be useful to view the results table for previous models?
* Would we want to show headers or some reminder of what the percentage is for clarification/context? Maybe we can add the "with `x`% accuracy" here?
Example: 
accuracy

* Since there's a lot of empty space, what if we filled it with the popup? 
Example: 
Screen Shot 2021-04-30 at 3 25 59 PM

* Do we want to prevent users from having duplicate models?
Screen Shot 2021-04-30 at 3 54 04 PM

breville:
Thanks for the reviews.  I'm going to merge this initial work, but will go through all the open comments for potential followups.

PR #160, opened 2021-04-26
Accuracy adjustments 

Erin007:
This PR includes two changes to improve accuracy and consistency of the trained models _with known datasets used in the curriculum_.  

1.) We are now training on the whole dataset 

2.) We set the K value as 1/3 of the dataset size 

Lengthy notes [here](https://docs.google.com/document/d/1Qi8A4JV5ymB74o2pfuAUo5fZpKp2Iu_FxZ-nQCUX0Og/edit#). With this change we get the expected accuracy for regression and classification models _using the toy datasets_  (I tested ice cream, pizza, salad, tacos, and safari) with single features and multiple features, regardless of order of feature selection. 

I see this as v1 of a more elaborate approach to using K value options to increase accuracy (in the vein of #161). However, I think this is worth merging the meantime while we build out a well-researched cohesive plan because it will immediately unblock the Curriculum team by addressing the accuracy inconsistencies seen in the datasets used in the curriculum.

### DIFF
--- a/src/train.js
+++ b/src/train.js
@@ -179,8 +179,8 @@ const prepareTrainingData = () => {
       numReserved++;
     }
   }
-  store.dispatch(setAccuracyCheckExamples(trainingExamples));
-  store.dispatch(setAccuracyCheckLabels(trainingLabels));
+  store.dispatch(setAccuracyCheckExamples(accuracyCheckExamples));
+  store.dispatch(setAccuracyCheckLabels(accuracyCheckLabels));
 };
 
 const prepareTestData = () => {
--- a/src/train.js
+++ b/src/train.js
@@ -147,8 +147,10 @@ const prepareTrainingData = () => {
   const trainingLabels = updatedState.data
     .map(row => extractLabel(updatedState, row))
     .filter(label => label !== undefined && label !== "" && !isNaN(label));
+  store.dispatch(setTrainingExamples(trainingExamples));
+  store.dispatch(setTrainingLabels(trainingLabels));
   /*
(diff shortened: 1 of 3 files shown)


dancodedotorg:
@Erin007 Clarifying question: will this change affect existing models?

Specifically: our Medical Priority model is specifically trained to get the results we want in that lesson to illustrate racial bias. If this changes the results of the already-trained model, then I'd need some time to redo that model.

(For further clarification: I'm not talking about making a new model on the existing dataset. I mean the model that's already been trained and already has an ID)

Erin007:
> @Erin007 Clarifying question: will this change affect existing models?
> 
> Specifically: our Medical Priority model is specifically trained to get the results we want in that lesson to illustrate racial bias. If this changes the results of the already-trained model, then I'd need some time to redo that model.
> 
> (For further clarification: I'm not talking about making a new model on the existing dataset. I mean the model that's already been trained and already has an ID)

This will not affect any trained models that already exist. 

 It _would_ mean that if you re-trained a model with the same dataset and the same selected features and label as existed before this change, you'd _potentially_ get a different accuracy score than the older model.

src/train.js
```diff
@@ -147,8 +147,10 @@ const prepareTrainingData = () => {
   const trainingLabels = updatedState.data
     .map(row => extractLabel(updatedState, row))
     .filter(label => label !== undefined && label !== "" && !isNaN(label));
+  store.dispatch(setTrainingExamples(trainingExamples));
+  store.dispatch(setTrainingLabels(trainingLabels));
```
breville, line 151: Maybe add a comment here that these are pre-cull?

breville (approved):

dancodedotorg (approved):

made-line (approved):

PR #162, opened 2021-04-27
use new isUserUploadedDataset prop to customize save model form 

Erin007:
When the user uploads a dataset `state.csvfile` will be an object; when it's an in-house curated dataset `state.csvfile` will be a string.  Here I use this difference as a shortcut to identify when the user has trained a model using an uploaded dataset so that we can customize the save model form by showing a text box to enter a description of the dataset and uncollapse the column description inputs. In cases where the trained model uses an in-house dataset, we show an uneditable description of the data and the column descriptions are hidden by default. This fixes the bug: "Describe the Data" textbox turns into to an uneditable block of text.

BEFORE 

https://user-images.githubusercontent.com/12300669/116280758-06064680-a757-11eb-9050-8e4f89e694c8.mov

AFTER 

https://user-images.githubusercontent.com/12300669/116280500-bf185100-a756-11eb-9dce-22506fabec60.mov

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -1293,3 +1293,9 @@ function areArraysEqual(array1, array2) {
     })
   );
 }
+
+export function isUserUploadedDataset(state) {
+  // The csvfile for internally curated datasets are strings; those uploaded by
+  // users are objects. Use data type as a proxy to know which case we're in.
+  return typeof state.csvfile === 'object' && state.csvfile !== null;
+}
--- a/src/UIComponents/SaveModel.jsx
+++ b/src/UIComponents/SaveModel.jsx
@@ -5,7 +5,8 @@ import { connect } from "react-redux";
 import {
   setTrainedModelDetail,
   getSelectedColumnDescriptions,
-  getDataDescription
+  getDataDescription,
+  isUserUploadedDataset
 } from "../redux";
 import { styles, saveMessages, ModelNameMaxLength } from "../constants";
 import Statement from "./Statement";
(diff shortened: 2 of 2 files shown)


src/redux.js
```diff
@@ -1293,3 +1293,9 @@ function areArraysEqual(array1, array2) {
     })
   );
 }
+
+export function isUserUploadedDataset(state) {
+  // The csvfile for internally curated datasets are strings; those uploaded by
+  // users are objects. Use data type as a proxy to know which case we're in.
+  return typeof state.csvfile === 'object' && state.csvfile !== null;
```
breville, line 1300: Overloading data type to have a deeper meaning, like this, makes me slightly nervous; it's often clearer to have an explicit value that represents how it was uploaded, especially as a way of future-proofing.  I guess the helper function does make things fairly clear, but I'm curious if you also have this concern...

breville:
As a side note, the carets don't show on `localhost`, though they do show when running in Code Studio.  We could adjust the way we load them to work everywhere.

breville (approved):

src/redux.js
Erin007: It makes me slightly nervous, too, which is why I added the comment to make it clearer that it's kind of hacky. Any time I find myself saying something is proxy, I usually mean it's a shortcut which has it's pros and cons 😄  To me this felt like the simpler and therefore safer choice at least in the short term. But I could be convinced to add a new piece of state,  that gets set in SelectDataset if you think this is a likely hole our future selves will stumble into.

breville: Let's go ahead with this for now, and we'll see how it feels later :)

PR #163, opened 2021-04-30
Fix: formate auto changes to align with style guide

BennettElisa:
Morning @Erin007!  

I moved to a new branch because I was unable to figure out how to align my previous branch to push the new changes related to formatting. This is the message I received in my terminal _"Your branch and 'origin/jest-test-classification' have diverged, and have 1 and 15 different commits each, respectively. (use "git pull" to merge the remote branch into yours)"._ I followed the instructions and did a git pull however I still keep running into the same message. If you have a moment to explain a solution to when this happens verses creating a new branch....not sure if this is considered bad practice or is acceptable?  I also want to continue to better understand the flow of pull request. When I create a new pull request related to a previous one on a different branch should I close out the other pull request so that this one is the active one? Thanks! -E


Erin007 (approved):

Erin007:
@BennettElisa 

Opening another PR with a different branch is fine, but probably not as efficient as it could've been since you had to re-do the changes you did want. There are a handful of ways to handle scenarios like this.  In my opinion the easiest would've been to run `git checkout main redux.js`. `git checkout <branch> <filename>` will restore the contents of that specific file to whatever it is on the specified branch. It this case, you'd have discarded all the changes your editor made to the redux.js file and it would no longer have any difference showing up in the comparison to the main branch. 

Yes, you should close the other pull request since you are actively using this branch and want to merge the changes in this PR. You can write a comment, something like "closing in favor of #163" and click the "Close with comment" button on the old PR. 

I general you'll want to get in the habit of pulling the main branch (or whatever branch you want to merge the change on your feature branch into) frequently and merging it into your feature branch so your feature branch doesn't become stale. 

Let me know if you'd like more clarification!

BennettElisa:
@Erin007 thanks for the feedback Erin! I've been keeping a handy reference guide and I've added your suggestion for what to do it the future as a an alternative to creating a new branch. I def. learned the lesson of pulling more frequently - from one day to the next there were 15 changes to the main branch!

PR #166, opened 2021-05-03
Continued iteration on K value optimization

Erin007:
Continuation of the work and thinking in #160

AI Lab uses the the KNN machine learning algorithm. KNN stands for K-Nearest Neighbors. The algorithm plots each feature vector in the training data. When a test vector is added to the plot, the algorithm calculates the distance between the test vector and its K nearest neighbors, then makes a prediction based on the classes or averaged numerical values of those neighbors. If K=3, then the 3 closest points of the training data to the test vector are considered when predicting the class or value of the test vector.  K is a hyper-parameter for the machine learning algorithm and setting its value to the optimal number of neighbors that will lead to the most accurate model is an important and hairy problem for data scientists. Different datasets and methods for determining the accuracy of a given model will have different optimal K values. 

Because we are using a variety of datasets, including allowing users to upload their own datasets, there is no single value of K that will be optimal for all potential datasets used to train models in AI Lab. To increase the odds of selecting a K value that will lead to as highly accurate a model as possible for a given dataset, we're trying out a handful of potential K values, calculating the accuracy for each of the resulting models and then using the most accurate model. 

I selected 1, 3, 5, 7, 17, 31, 45, 61, the square root of N where N is the dataset size, and 1/3 dataset size for the potential values of K to try. These numbers were selected based on this [academic paper](https://arxiv.org/pdf/1409.0919.pdf) and trial and error for the current in-house datasets used in the curriculum (additional [notes](https://docs.google.com/document/d/1Qi8A4JV5ymB74o2pfuAUo5fZpKp2Iu_FxZ-nQCUX0Og/edit)). 

This limited list of values does not guarantee that the optimal value of K will be found for all possible datasets; to do so would require testing each possible value of K, which is time and computationally intensive. For our purposes, the goal isn't to perfectly optimize accuracy for every possible combination of features and label selection for every possible dataset. Our goal is to demonstrate that AI can be trained to identify known patterns in datasets straightforward enough as to be approachable for a middle school audience. To that end, this implementation seems a good starting point for optimizing accuracy for trained models we're likely to encounter based on the AI unit pilot currently underway.

How are we going to know whether the list of potential values I picked are good ones? 
- We can do some analysis of accuracy scores and K values once we have a solid set of student-trained models to see if there are any patterns. For example, we could query the K values for all trained models and see if any of the potential K values are _not_ being used and remove them for the list of K values to try. We could see if there is a strong correlation between dataset size and the K value for the most accurate model. Or number of features and the K value for the most accurate model, or classification v. regression patterns and adjust the list accordingly. 
- We could ask users to report scenarios where they know there is a strong pattern to find in a dataset, but where AI Lab failed to generate an accurate model and run additional K values for that dataset/set of features and see if there is an additional K value heuristic we should try. 
- We could run an internal study of all feature + label selection combos for a variety of internally curated datasets and all possible Ks and cross reference the optimal Ks with the list we're trying to see if it needs to be adjusted. 
- other ideas... 

Mostly noting these as reminders to our future selves while it's fresh for a later date if we find this issue warrants further investigation. I don't plan on pursuing additional work on optimizing K until we have evidence that we need to; however, I did add K to the data that gets saved when a model is saved so we have it to reference for future analysis if needed.

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -57,6 +57,7 @@ const SET_HIGHLIGHT_DATASET = "SET_HIGHLIGHT_DATASET";
 const SET_RESULTS_PHASE = "SET_RESULTS_PHASE";
 const SET_INSTRUCTIONS_KEY_CALLBACK = "SET_INSTRUCTIONS_KEY_CALLBACK";
 const SET_SAVE_STATUS = "SET_SAVE_STATUS";
+const SET_K_VALUE = "SET_K_VALUE";
 
 // Action creators
 export function setMode(mode) {
@@ -200,6 +201,10 @@ export function setSaveStatus(status) {
   return { type: SET_SAVE_STATUS, status };
 }
 
+export function setKValue(kValue) {
+  return { type: SET_K_VALUE, kValue };
+}
+
 const initialState = {
   name: undefined,
   csvfile: undefined,
@@ -230,7 +235,8 @@ const initialState = {
   currentColumn: undefined,
   resultsPhase: undefined,
   saveStatus: undefined,
-  columnRefs: {}
+  columnRefs: {},
+  kValue: null
 };
 
 // Reducer
@@ -523,6 +529,12 @@ export default function rootReducer(state = initialState, action) {
       saveStatus: action.status
     };
   }
+  if (action.type === SET_K_VALUE) {
+    return {
+      ...state,
+      kValue: action.kValue
+    };
+  }
   return state;
 }
 
@@ -737,7 +749,7 @@ export function getConvertedPredictedLabel(state) {
   return getConvertedLabel(state, state.prediction.predictedLabel);
 }
 
-export function getConvertedLabels(state, rawLabels) {
+export function getConvertedLabels(state, rawLabels = []) {
   return rawLabels.map(label => getConvertedLabel(state, label));
 }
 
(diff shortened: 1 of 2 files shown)


Erin007:
@moneppo FYI in case you have thoughts on this as you look into continuing to curate our in-house datasets.

Erin007:
A couple additional notes: 
1.) I didn't see any reduction in performance with this additional step of trying multiple K values. We've time built in while users are on the training screen watching the animation that is more than sufficient to try these K values out without compromising the flow, but it's something to watch out for with this change. 
2.) The punchline of the academic paper I linked is actually that the authors have devised an ensemble learning approach that outperforms KNN, but I'm hoping to avoid the complexity and steep learning curve of going into that tunnel unless we really think we have to.  I was inspired by their methodology. 
3.) I didn't do any optimizations around de-duping the list of potential Ks (if for example 1/3 of dataset size happens to match one of the potential K values already in the list), since it seemed both unlikely and not cost-intensive to run the dupe; but worth noting if we really want to dig into performance. 
 4.) I didn't bother with distinguishing between ties (e.g. if  k=5 and k=7 have the same best accuracy scores, we'll use k=5) which means we're not going to be able to know much about ties in any follow up analysis.

breville (approved):

Erin007:
https://docs.google.com/document/d/11u2ZZ-5Ols8FECsBBw6HdmWyKpBjwO5uwPBeaw80dRg/edit

PR #167, opened 2021-05-06
Adjust error tolerance for regression

Erin007:
Previously, regression predictions were considered correct if they were within +/- 3% of the expected value. In this PR, I adjusted that tolerance range to +/- 5%.  I also extracted the tolerance into a variable so if we need to change it again, we can make that change in one place instead of two. 

update error tolerance

### DIFF
--- a/src/UIComponents/ResultsTable.jsx
+++ b/src/UIComponents/ResultsTable.jsx
@@ -8,7 +8,12 @@ import {
   getAccuracyGrades,
   isRegression
 } from "../redux";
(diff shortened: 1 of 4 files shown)


Erin007:
@moneppo heads up for your dataset work that this change is incoming. Let me know if it resolves the scenarios you were expecting it to or not. 

Also, tried to tag you, but looks like your invite to the repo is still pending? Join the party!

breville (approved):

PR #168, opened 2021-05-06
Levelbuilders can set randomizeTestData

Erin007:
This change re-introduces the redux actions to setReserveLocation which customizes whether the test data come from the beginning or end of the dataset or is randomly selected. This was reintroduced to allow levelbuilders to add `randomizeTestData = true` in the mode JSON for a level.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,7 +10,8 @@ import rootReducer, {
   setSelectedJSON,
   setSelectedTrainer,
   setInstructionsKeyCallback,
-  setSaveStatus
+  setSaveStatus,
+  setReserveLocation
 } from "./redux";
 import { allDatasets } from "./datasetManifest";
 import { parseCSV } from "./csvReaderWrapper";
--- a/src/index.js
+++ b/src/index.js
@@ -15,6 +15,7 @@ import rootReducer, {
 import { allDatasets } from "./datasetManifest";
 import { parseCSV } from "./csvReaderWrapper";
 import { parseJSON } from "./jsonReaderWrapper";
+import { TestDataLocations } from "./constants";
 
 export const store = createStore(rootReducer);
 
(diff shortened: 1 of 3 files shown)


Erin007:
Note: we currently have a similar mode `hideChooseReserve` the intent of which is to show or hide a dropdown that will allow the user to customize where the test data comes from. Right now we're not ever showing this dropdown to users, and the code for this mode is ripe for deletion unless we decide there's a need to show users this dropdown. 
Screen Shot 2021-05-06 at 12 13 52 PM

Similarly, we aren't currently ever pulling test data from the beginning of the dataset, so if we decide to fully remove the dropdown selector and hideChooseReserve mode we can remove the code related to selecting from the beginning of the dataset.

src/index.js
```diff
@@ -74,6 +76,10 @@ const processMode = mode => {
     if (mode.hideSelectTrainer) {
       store.dispatch(setSelectedTrainer(mode.hideSelectTrainer));
     }
+
+    if (mode.randomizeTestData) {
```
breville, line 80: Do we think we'll want to offer more than just random at some point?

Erin007: Potentially in the far future, but not in the current unit.

breville: My one thought was whether we wanted to expose this parameter a bit more generally, but if you're comfortable only exposing randomization, then that seems fine.

breville (approved):

src/index.js
Erin007: Yeah, I thought about that too and feel like we can add that if we need it or roll it into levelbuilder improvements; I suspect form fields rather than a JSON blob of parameters will be a nicer long-term interface once the parameters are completely solidified.

Erin007:
cc: @dancodedotorg once this change hits levelbuilder (planning on Monday) you can set randomizeTestData = true in the JSON for the level and the test data will be pulled randomly rather than from the end.

PR #170, opened 2021-05-07
Delete code related to trainer selection

Erin007:
Previous iterations of the tool included an additional algorithm (SVM) and a feature where users could select which algorithm they wanted to use. Now that we are exclusively using KNN, and inferring classification or regression based on label column data type (numerical labels use regression, categorical columns use classification) we can remove the code related to selecting a trainer.  

This PR includes: 

- Removing references to the levelbuilder param `hideSelectTrainer` because the trainer selection is no long available so the param is no longer needed. 
- Removing the `selectedTrainer` piece of state and related reducer action 
- Removing the `getMLType` helper function because it is redundant with `isRegression` which is now based on data type of label column
- Removing the `availableTrainers` object, which is not used outside of the newly deleted `getMLType` function
- Renaming `deafultRegressionTrainer` to `RegressionTrainer` (similar for classification) and moving it to the constants.js file both to avoid and import from train.js into redux.js and because default implies the default of many possible, but we just have one regression trainer and one classification trainer.

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -1,10 +1,4 @@
 import {
-  getMLType,
-  defaultRegressionTrainer,
-  defaultClassificationTrainer
-} from "./train.js";
-
-import {
   datasetUploaded,
   uniqueColumnNames,
   noEmptyCells,
@@ -20,6 +14,8 @@ import {
 import {
   ColumnTypes,
   MLTypes,
+  RegressionTrainer,
+  ClassificationTrainer,
   TestDataLocations,
   ResultsGrades,
   REGRESSION_ERROR_TOLERANCE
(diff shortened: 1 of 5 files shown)


src/redux.js
```diff
@@ -972,7 +938,10 @@ export function getTrainedModelDataToSave(state) {
   dataToSave.potentialUses = state.trainedModelDetails.potentialUses;
   dataToSave.potentialMisuses = state.trainedModelDetails.potentialMisuses;
 
-  dataToSave.selectedTrainer = getSelectedTrainer(state);
+  dataToSave.selectedTrainer =
+    isRegression(state)
+    ? RegressionTrainer
+    : ClassificationTrainer;
```
made-line, line 944: tiny nit: should we indent these lines?

made-line (approved):
LGTM!

breville (approved):

PR #177, opened 2021-05-12
Use Dynamic Instructions overlay

breville:
https://github.com/code-dot-org/code-dot-org/pull/40469 adds support for showing the overlay to dim the rest of the screen when Dynamic Instructions are shown or changed.  With this change, AI Lab will display that overlay each time a scene is shown for the first time.  It also handles a notification from the instructions when the overlay has been dismissed, so that we can pause the action until the overlay is gone; this is demonstrated by not starting the `TrainModel` animation until the overlay is gone.

Level editors can disable this behaviour for a level by setting the `hideInstructionsOverlay` parameter to `true`, as is demonstrated by the `preload-metadata` sample mode.

When running this app standalone, clicking the instruction ID at the top of the screen will simulate dismissing the overlay.

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -55,6 +55,7 @@ const SET_RESULTS_PHASE = "SET_RESULTS_PHASE";
 const SET_INSTRUCTIONS_KEY_CALLBACK = "SET_INSTRUCTIONS_KEY_CALLBACK";
 const SET_SAVE_STATUS = "SET_SAVE_STATUS";
 const SET_K_VALUE = "SET_K_VALUE";
+const SET_INSTRUCTIONS_DISMISSED = "SET_INSTRUCTIONS_DISMISSED";
 
 // Action creators
 export function setMode(mode) {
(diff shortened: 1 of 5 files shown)


src/redux.js
```diff
@@ -342,11 +349,11 @@ export default function rootReducer(state = initialState, action) {
     };
   }
   if (action.type === SET_RESERVE_LOCATION) {
```
Erin007, line 351: The changes below look like formatting tweaks? Are my editors settings not in sync with yours?

Erin007 (approved):

src/redux.js
breville: Possibly JsPrettier just wasn't run on a particular commit?  It does seem that the spacing was off by one when committed [here](https://github.com/code-dot-org/ml-playground/commit/8058751c41ae869dbbf4b81d1e31fd6999349fef#diff-7f797aa32260ac4cbc8b43b04d62976c18e503bea6cc86be7448f935896ec3cfR362-R367).

PR #178, opened 2021-05-12
Delete training settings

Erin007:
Previous iterations of the tool included additional training customizations (selecting between algorithms, setting the K value, etc) that were shown on a pre-training screen. In the latest version of the tool, that screen only shows dropdowns to confirm data type for selected columns, which is redundant to functionality in the column inspector side panels. Thus I am removing the training screen, related components and residual code. Additionally, we no longer use the option to select test data from the beginning of the dataset, so deleting that code here too.


Erin007:
@dancodedotorg with this change the  `hideSpecifyColumns` and `hideChooseReserve` parameters will no longer be needed.  If the user has selected an in-house dataset, we'll use the column datatypes from the JSON metadata; if it's a user-uploaded dataset, we default to categorical and show a selection dropdown in the column inspector side panel. Test data reservation percentage is now always set to 10%. Thanks!

breville (approved):

PR #179, opened 2021-05-12
Adding four new real world datasets

moneppo:


public/datasets/basketball_shots.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    {
+      "type": "numerical",
+      "id": "Time in seconds",
+      "description": "Time elasped on the game clock, in seconds."
+    },
+    {
+      "type": "numerical",
+      "id": "Jersey number",
+      "description": "The number of the player who took the shot."
```
Erin007, line 32: Hmm - this one is interesting. It's a number, but not one that would make sense on a number line. Like you wouldn't want to use this column for regression, but I think it's fine. Just noting. 🤷‍♀️

public/datasets/jeans.json
```diff
@@ -0,0 +1,76 @@
...
+      },
+      {
+        "type": "numerical",
+        "id": "Max width of back pocket",
+        "description": "The width of the front pocket from its widest part."
+      },
+      {
+        "type": "numerical",
+        "id": "Min width of back pocket",
+        "description": "The width of the front pocket from its most narrowest part."
```
Erin007, line 62: maybe "most narrow" or  just "narrowest"?

src/datasetManifest.js
```diff
@@ -1,11 +1,46 @@
...
+    path: "datasets/exoplanets.csv",
+    metadataPath: "datasets/exoplanets.json",
+    imagePath: "datasets/exoplanets.jpg"
+  },
+  {
+    id: "thanksgiving",
+    name: "Regional Thanksgiving Food Survey",
+    path: "datasets/thanksgiving.csv",
+    metadataPath: "datasets/thanksgiving.json",
+    imagePath: "datasets/thanksgiving.jpg"
```
Erin007, line 35: Is this PR missing the Thanksgiving JSON file?

public/datasets/thanksgiving.json
```diff
@@ -0,0 +1,205 @@
+{
+  "card": {
+    "description": "A survey from FiveThrityEight, asking people around the country what foods they have for Thanksgiving dinner.",
```
Erin007, line 3: maybe "asking people across the United States"? It's a US-centric dataset and I'm being nit-picky but know we want to think in terms of a global audience as we continue to expand worldwide usage?

Erin007 (approved):

PR #180, opened 2021-05-14
Generate Results & other UI

breville:
This change contains the addition of `GenerateResults`, some improvements to `TrainModel`, and some other proposed UI changes.

`GenerateResults` shows the A.I. bot making its predictions which are then used to calculate accuracy.  The data shown is actually from the source data table, making it more illustrative than strictly accurate, but this can be improved if desired.

`TrainModel` and `GenerateResults` now show values being "plucked" out of the table, rather than scrolling the table, in an attempt to make it clearer that the box that moves across the screen is a single row from the source table.  Fades are used in an attempt to make it clearer that the data processing shown is just a subset of the overall process, which might involve hundreds or thousands of rows.

Some other UI work is presented here for consideration:

- The `DataCard` no longer shows all the column names and descriptions.  We received feedback that this information was overwhelming on the first screen, and the information can be discovered on the subsequent screens anyway.
- The `DataTable` now only shows the first 100 rows of data, with a small message explaining if this is the case.  This significantly improves UI performance, since the UI code was never optimised to cope with hundreds or thousands of rows of data.  All data processing continues to use the full data sets.
- `TrainModel` can now be skipped immediately.  The three second delay was arbitrary; we now have the overlay to slow the viewer down, and the animation has a fixed duration for viewers who want to keep watching.
- Instead of the `Statement`, `TrainModel` and `GenerateResults` now show specific headings to draw attention to what these screens are doing.
- `TrainModel` no longer has text in its background image, which would have been difficult to localise.

https://user-images.githubusercontent.com/2205926/118765520-0a8ebe00-b830-11eb-9544-762b1fe66b5f.mov


dancodedotorg (approved):

src/UIComponents/TrainModel.jsx
```diff
@@ -23,11 +23,16 @@ class TrainModel extends Component {
...
-    animationTimer: undefined,
-    headOpen: false
-  };
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      frame: 0,
+      animationtimer: undefined,
+      headOpen: false,
```
Erin007, line 32: I love this variable name 😆

Erin007:
- When I run through the flow once, my results show twice - once as the top result and then in the previous results section. I think of previous as exclusive of current. Was that intentional? 
Screen Shot 2021-05-24 at 10 05 49 AM

Doesn't need to be part of this PR, but noting for the next style pass...
- I'm a big fan of the bot head closing on the training screen, gives a nice sense of completion to the process! 
- Wondering about the "Testing Accuracy" heading, maybe "Calculating Accuracy" or "Testing the model" ? 
- On the label selection screen, what do we think about truncating the statement to just "Predict..." then adding "based on..." when we get to feature selection? Having lingering ... makes it feel like I haven't done everything I need to on that screen, but I have. 
- The bot jumps up as we transition from training to testing. Maybe it's fine, but wondered if the bot could be positioned similarly on both screens?  🤷‍♀️  Or maybe the testing bot doesn't appear until the instructions are dismissed? 
- Maybe another Bot opportunity in the "Try it out!" section. Instead of the little grey box, we could re-use the testing beam?
Screen Shot 2021-05-24 at 10 08 03 AM
- I don't think we need pixel-perfect consistency between AI Lab and Oceans, but noticed that the beam strength and width differs. My only real gripe here is that I think we could make the beam color stand out a nudge more on the gray background. 
Screen Shot 2021-05-24 at 10 12 23 AM
Screen Shot 2021-05-24 at 10 10 24 AM

Erin007:
I tried the flow with many features selected, and it worked pretty well: 
Screen Shot 2021-05-24 at 10 20 24 AM
Screen Shot 2021-05-24 at 10 20 15 AM

Except things get a little squishy on the results table, which we could tidy up a bit. 
Screen Shot 2021-05-24 at 10 20 01 AM

I still had a pretty wide screen though; long rows could be cut off on smaller screens. I know @moneppo was looking at mobile, so wanted to note that we should check models with many features on smaller screens as part of that.

Erin007:
Screen Shot 2021-05-24 at 10 26 02 AM
Now that we removed all of the column options from the DataCard (which feels like a good call!) I think we could bump the font size - feels tiny right now.

Erin007 (commented):
Yay! Great work, here! No blockers that I see in the code. I added a handful of suggestions around UI improvements.  The only thing that felt like a potential issue is that a single run through shows up in both the current and previous sections of the results table.

dancodedotorg:
> Wondering about the "Testing Accuracy" heading, maybe "Calculating Accuracy" or "Testing the model" ?
> * On the label selection screen, what do we think about truncating the statement to just "Predict..." then adding "based on..." when we get to feature selection? Having lingering ... makes it feel like I haven't done everything I need to on that screen, but I have.
> * Maybe another Bot opportunity in the "Try it out!" section. Instead of the little grey box, we could re-use the testing beam?

Just wanted to +1 each of these. I like "Testing the Model" vs "Testing Accuracy" - that feels easier for a teacher or student to explain over-the-shoulder if a student is confused on what's happening.

Also, adding AI Bot to the "Try it out!" feels like a good move. We got lots of feedback that having AI Bot be more visible in the process felt good to students, so this seems to meet that feedback.

breville:
Thanks @Erin007 and @dancodedotorg.  I've fixed the double generation of results, and will review all the UI feedback next since that's my focus for the next two weeks.

made-line:
Looks great! I like how the animation on the TrainModel screen has a fixed duration. I also appreciate the improved UI performance!

PR #184, opened 2021-05-17
Restrict label and feature selection to columns with less than 50 unique values

Erin007:
In this PR I restrict label and feature selection to categorical columns with less than 50 unique values; no restrictions for numerical columns. Rather than a "Select label" or "Select feature" button, the user will now see a message explaining why that column can not be selected. 

limit unique values

This is to alleviate to related issues, one technical and one curricular/product. 

1.) The metadata column of the UserMLModels table is a text field that can hold 65,535 characters. In the metadata field we store all the possible values for selected features and label so we can display that information in the Model Card in App Lab and as values in the autogenerated test input dropdowns. Features with many possible values - such as predicting budget based on director and start with the movies dataset exceed the maximum limit for metadata size. 

2.) Categorical columns with a many unique options don't provide much meaningful information to this type of machine learning algorithm and thus create very inaccurate models. Early classroom tests revealed students wanted to try to predict columns such as pokemon name - all unique values - which is more consistent with ideas about "look up" rather than "classification"; explicitly preventing students from selecting label or feature columns with too many unique options constrains users to scenarios that are more aligned with the task at hand.

### DIFF
--- a/src/UIComponents/ColumnInspector.jsx
+++ b/src/UIComponents/ColumnInspector.jsx
@@ -11,7 +11,7 @@ import {
   getRangesByColumn,
   setCurrentColumn
 } from "../redux";
-import { ColumnTypes, styles } from "../constants.js";
+import { ColumnTypes, styles, UNIQUE_OPTIONS_MAX } from "../constants.js";
 import { Bar } from "react-chartjs-2";
 import ScatterPlot from "./ScatterPlot";
 import CrossTab from "./CrossTab";
@@ -214,7 +214,8 @@ class ColumnInspector extends Component {
               </form>
 
               {currentPanel === "dataDisplayLabel" &&
-                currentColumnData.id !== labelColumn && (
+                currentColumnData.id !== labelColumn &&
+                !currentColumnData.hasTooManyUniqueOptions && (
                   <button
                     type="button"
                     onClick={e =>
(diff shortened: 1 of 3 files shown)


src/constants.js
```diff
@@ -27,6 +27,8 @@ export const TestDataLocations = {
 
 export const ModelNameMaxLength = 150;
 
+export const UNIQUE_OPTIONS_MAX = 50;
```
Erin007, line 30: 50 is kind of arbitrary - it's smaller than it would need to be to alleviate the technical concern and still large enough that it doesn't guarantee high model accuracy.  I think it probably warrants a conversation about how restrictive we'd like to be from the curricular/product perspective?

Erin007: @moneppo how does this cut-off number feel as related to the new datasets you're finding?

dancodedotorg: Clarifying that this would only affect *categorical* data, yes? For example, numerical would definitely have more than 50 unique values, but that's expected since we want to train on the range.

If this is only affecting categorical data, then 50 options max feels fine to me curricularly. I'd defer to Mike on what he's observed in the datasets he's been collecting if that feels off from his experiences.

dancodedotorg: D'oh! I just saw the screenshot that says "categorical columns...". Please disregard my question above =P

src/UIComponents/ColumnInspector.jsx
```diff
@@ -236,6 +238,14 @@ class ColumnInspector extends Component {
                     Add feature
                   </button>
                 )}
+
+                {currentColumnData.hasTooManyUniqueOptions && (
+                  <span>
+                    Categorical columns with more than {UNIQUE_OPTIONS_MAX}
+                    {" "} unique values can not be selected as the label or
+                    {" "} features.
```
breville, line 246: Possibly "a feature"?

breville (approved):

PR #187, opened 2021-05-18
Refactor to use `getExtrema` 

Erin007:
Refactoring a bit of Redux code to better align with best practices in "Clean Code": 

- remove a flag argument 
- keep functions small and their purpose singular 
- provide a more descriptive function name

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -643,7 +643,7 @@ export function getCurrentColumnData(state) {
     readOnly: isColumnReadOnly(state, state.currentColumn),
     dataType: state.columnsByDataType[state.currentColumn],
     uniqueOptions: getUniqueOptions(state, state.currentColumn),
-    range: getExtrema(state, state.currentColumn),
+    extrema: getExtrema(state, state.currentColumn),
     frequencies: getOptionFrequencies(state, state.currentColumn),
     description: getColumnDescription(state, state.currentColumn),
     hasTooManyUniqueOptions: hasTooManyUniqueOptions(state, state.currentColumn)
@@ -1011,9 +1011,9 @@ export function getColumnDataToSave(state, column) {
   if (state.columnsByDataType[column] === ColumnTypes.CATEGORICAL) {
     columnData.values = getUniqueOptions(state, column);
   } else if (state.columnsByDataType[column] === ColumnTypes.NUMERICAL) {
-    const maxMin = getExtrema(state, column);
-    columnData.max = maxMin.max;
-    columnData.min = maxMin.min;
+    const {max, min} = getExtrema(state, column);
+    columnData.max = max;
+    columnData.min = min;
   }
   return columnData;
 }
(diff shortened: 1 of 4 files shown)


src/redux.js
```diff
@@ -991,7 +999,7 @@ export function getColumnDataToSave(state, column) {
   if (state.columnsByDataType[column] === ColumnTypes.CATEGORICAL) {
     columnData.values = getUniqueOptions(state, column);
   } else if (state.columnsByDataType[column] === ColumnTypes.NUMERICAL) {
-    const maxMin = getRange(state, column, false);
+    const maxMin = getExtremums(state, column);
```
maureensturgeon, line 1002: Could also be written as `const {max, min} = getExtremums(state, column);`

src/redux.js
```diff
@@ -721,24 +721,33 @@ export function getUniqueOptionsByColumn(state) {
...
-  }
+function getMinimumValue(state, column) {
+  return Math.min(...state.data.map(row => parseFloat(row[column])));
+}
+
+function getRange(maximumValue, minimumValue) {
+  return Math.abs(maximumValue - minimumValue);
+}
+
+export function getExtremums(state, column) {
```
maureensturgeon, line 744: So readable!

maureensturgeon (approved):
So awesome! Thanks for sharing :)

cforkish:
i love this idea of practicing what we're reading about!

naming is a subjective art and you have context about this code that i do not, so please feel free to ignore my opinions about your naming choices. but since this PR is presented in the context of our book club, i feel like some friendly discussion can't hurt!

to my eye, `getExtremums` is actually _less_ clear than `getRange`.

for one, "extremums" is just a harder word to process with a less obvious meaning than "range". also i'm not positive it's a word -- according to [Britannica](https://www.britannica.com/science/extremum), the plural of "extremum" is "extrema".

but more to the point, it seems like an unnecessarily vague term to use to describe the extrema of a 1D column (if i'm understanding its use correctly, which maybe i'm not). a column has an explicit set of extrema -- the min and max -- so using a term like "range" which implies that same specific set seems clearer than "extremums", which to me implies a more arbitrary set of values.

put another way, thinking about it as a property of a `column`, `column.range` is much easier for me to parse than `column.extremums`. and `column.extremums.range` just seems unnecessarily obtuse.

that said, i don't love the previous definition of `range` either -- if i were making a `range` object, it would be `{start, end, length}` instead of `{min, max, range}`.

if you want to stick with `extremums` i think `{min, max, range}` makes sense. however, i would _much_ prefer `extrema` over `extremums` because the latter makes my grammar hurt.

regardless of what you decide to roll with for names, it would be great if you could add some comments explaining what they mean in this context.

like i said, take all that with a grain of salt. overall i very much approve of this effort!

src/redux.js
```diff
@@ -721,24 +721,33 @@ export function getUniqueOptionsByColumn(state) {
   return uniqueOptionsByColumn;
 }
 
-export function getRangesByColumn(state) {
-  let rangesByColumn = {};
+export function getExtremumsByColumn(state) {
```
cforkish, line 724: could you add a block comment explaining what this is for?

src/redux.js
cforkish, line 744: could you add a block comment explaining what the `extremums` return value means?

imo it would be worthwhile to actually define an `extremumsType` `PropType` so you can validate `Predict.propTypes.extremumsByColumn`

made-line (approved):
LGTM. Good call on removing the argument flag for getRange!

PR #191, opened 2021-05-25
Dataset tiny tidy for Thanksgiving and NCAA datasets 

Erin007:
There were a [few reports](https://codedotorg.slack.com/archives/C016VGH1BT6/p1621557040014100) of column data not appearing when certain columns were clicked. The problem occurred because in order to show the column information and graph we need to know the datatype for the column. In the code we use the column heading from the dataset's csv to do a look-up in the metadata from the JSON file to find type.  If the column heading from the csv doesn't exactly match the JSON metadata, we hit the error. 

Screen Shot 2021-05-24 at 10 57 26 PM

In the Thanksgiving dataset, there were typos in the cranberry, sweet potato pie and cheesecake columns which I corrected to alleviate the issue. For example:
Screen Shot 2021-05-24 at 10 14 07 PM

Similarly, an additional space in the basketball csv was causing the error for the Made or miss? column. I removed the space. 
Screen Shot 2021-05-24 at 10 52 23 PM


Erin007:
@moneppo this error occurs with many of the columns in the student alcohol consumption dataset. The csv uses abbreviations (e.g. "Walc" and "famrel") whereas the JSON uses full words (e.g. "weekend alcohol" and "family relationships"). I don't know where we stand on this dataset, so I'll let you decide to clean or cut. Thanks!

src/UIComponents/GenerateResults.jsx
```diff
@@ -2,7 +2,6 @@
 import PropTypes from "prop-types";
 import React, { Component } from "react";
 import { connect } from "react-redux";
-import train from "../train";
```
Erin007, line 5: @breville this and below were unrelated changes I needed to make the linter happy. Did something unexpected sneak in somewhere? 🤷‍♀️

breville: Thanks for fixing this.  I missed a failing check for the final commit to https://github.com/code-dot-org/ml-playground/pull/180.

breville (approved):

made-line (approved):

made-line:
Tiny nit: this may not be in the scope of this PR, but do we want to keep the question marks in the label/feature names with binary data? It doesn't match what's seen in other datasets (e.g., Zoo, Student Health Survey, Bike Sharing, etc.). It also looks a little random with the question mark: 
Screen Shot 2021-05-25 at 12 37 56 AM
@moneppo

dancodedotorg:
> Tiny nit: this may not be in the scope of this PR, but do we want to keep the question marks in the label/feature names with binary data? It doesn't match what's seen in other datasets (e.g., Zoo, Student Health Survey, Bike Sharing, etc.). It also looks a little random with the question mark:

Ooo - good call out Madline. I'll post about this in #ai-lab - we need to change this to match the other datasets.

PR #194, opened 2021-05-26
Save dataset name for metrics

Erin007:
We'd like to be able to track which dataset was used to train a saved model for metrics. There are a handful of gnarly ways we could've inferred which in-house dataset was used from the metadata we were already saving with each trained model, but this is felt much cleaner and easier to query. I added name to the datasetDetails object already included in the data we save.

### DIFF
--- a/public/datasets/exoplanets.json
+++ b/public/datasets/exoplanets.json
@@ -1,71 +1,71 @@
 {
-    "card": {
-      "description": "This dataset is a list of exoplanets, which are planets orbiting other stars. With this data, it's possible to estimate the size of an exoplanet as well as other characteristics.",
-      "source": "https://exoplanetarchive.ipac.caltech.edu/",
-      "lastUpdated": "2021",
-      "context": {
-        "createdBy": "California Institute of Technology",
-        "potentialUses": "Estimate the size or mass of an exoplanet based on the size, age or other aspects of the star it orbits.",
-        "potentialMisuses": "Trying to make guesses as to whether a planet could support life."
-      }
+  "name": "exoplanets",
+  "card": {
+    "description": "This dataset is a list of exoplanets, which are planets orbiting other stars. With this data, it's possible to estimate the size of an exoplanet as well as other characteristics.",
+    "source": "https://exoplanetarchive.ipac.caltech.edu/",
+    "lastUpdated": "2021",
+    "context": {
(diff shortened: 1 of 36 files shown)


Erin007:
@moneppo Can you add names to the JSON files for any new datasets you add, please? And ensure that the JSONifier tool we inherited from Brent also adds "name"? Thanks!

public/datasets/exoplanets.json
```diff
@@ -1,71 +1,71 @@
...
-      "description": "This dataset is a list of exoplanets, which are planets orbiting other stars. With this data, it's possible to estimate the size of an exoplanet as well as other characteristics.",
-      "source": "https://exoplanetarchive.ipac.caltech.edu/",
-      "lastUpdated": "2021",
-      "context": {
-        "createdBy": "California Institute of Technology",
-        "potentialUses": "Estimate the size or mass of an exoplanet based on the size, age or other aspects of the star it orbits.",
-        "potentialMisuses": "Trying to make guesses as to whether a planet could support life."
-      }
+  "name": "exoplanets",
+  "card": {
```
Erin007, line 3: This looks like a big diff, but I just fixed misaligned indentation. Same for jeans and songs.

breville: Looks good with https://github.com/code-dot-org/ml-playground/pull/194/files?diff=unified&w=1

breville (approved):

made-line (approved):
LGTM!

src/redux.js
```diff
@@ -987,6 +987,7 @@ export function getDataDescription(state) {
 
 export function getDatasetDetails(state) {
   const datasetDetails = {};
+  datasetDetails.name = state.metadata.name;
```
breville, line 990: Oh, what happens for user-uploaded CSVs?  Do we want to save a name for them too?

Erin007: Great question. I asked @moneppo about it, and it's not a priority for this go-around. We'd like to know is the dataset user uploaded or not and if it is not, which of the in-house datasets was selected. Based on that metric (and curriculum feedback etc) if user uploaded datasets are common we can invest in having students name their datasets and maybe move some of the data collection questions currently on the model save form earlier in the flow to right after upload. For now, though, this is all we need :)

PR #195, opened 2021-05-27
Fix Thanksgiving dataset bug and update Thanksgiving/NCAA formatting

made-line:
This PR addresses the following:
* When using the Thanksgiving dataset, certain label/feature combinations made the animation on the Training screen freeze. The current workaround is to remove rows with empty cells. See my [notes](https://docs.google.com/document/d/1bGKrH5DNfjotzfgvCO3Z4af-mVLLbiBL-SY7gUOu7GU/edit?usp=sharing) for more details. 
* The question marks are removed from both the Thanksgiving and NCAA datasets (per https://github.com/code-dot-org/ml-playground/pull/191). Thank you for your help, @moneppo!

Before: 

https://user-images.githubusercontent.com/40412372/119754358-396eea80-be55-11eb-8975-c5d974b062c4.mov

After: 

https://user-images.githubusercontent.com/40412372/119754421-54d9f580-be55-11eb-9505-46f604943d44.mov

### DIFF
--- a/public/datasets/basketball_shots.csv
+++ b/public/datasets/basketball_shots.csv
@@ -1,4 +1,4 @@
-Away team alias,Home team alias,Time in seconds,Jersey number,Inches from left baseline,Inches from bottom sideline,Made or miss?
+Away team alias,Home team alias,Time in seconds,Jersey number,Inches from left baseline,Inches from bottom sideline,Made or miss
 FSU,MIZZ,556,10,373,335,miss
 FSU,MIZZ,808,15,282,459,miss
 FSU,MIZZ,379,15,339,207,miss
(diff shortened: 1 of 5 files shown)


breville (approved):

dancodedotorg (approved):

Erin007:
@moneppo are you handling removing empties from other datasets? I'm worried about the user-upload scenario since I don't think we can trust students not to have empty cells. Are we tracking a programmatic way to either 1.) notify students when their uploaded CSVs have empty cells 2.) auto-remove those rows for them or 3.) modify the code to cope with empty cells?

Erin007 (approved):

moneppo:
Yes, I have just ran a pass over all datasets and removed empty rows.

As for the student-uploaded case, I've enumerated the rules I think we need
to add here:
https://trello.com/c/rMFXOnqT/223-update-validation-for-empty-cells-in-uploaded-csvs

PR #196, opened 2021-05-27
UI fit & finish

breville:
Lots of UI fit & finish across the whole AI Lab app, including:

- Horizontal lines to separate panel headings from bodies.
- Dark grey backgrounds are gone.
- Font sizes variations are reduced, it's mostly 24px and 14px now.
- Upload CSV is reduced to a single button.
- User can now upload a CSV, then choose a built-in dataset, and then re-upload the same CSV and it will be used.
- Force wrapping on long strings (e.g. source URLs) to prevent horizontal scrolling in right panel.
- Dataset tiles can be various heights due to different title lengths and still lay out cleanly.
- Statement font size is reduced.
- Statement gets underscores for blank fields.
- Statement just shows "Predict _____" at first, then "based on _____" when ready to add features.
- Removed X in column inspector.
- Column inspector fields get clear headings.
- Ensured buttons to select label & features are always visible, scrolling content above if necessary.
- GenerateResults heading changed to "Testing the model".
- The A.I. bot now slides from the middle to the top in GenerateResults.
- The predict beam in GenerateResults is now more saturated and is correctly positioned against the bot.
- Subtle hover highlight for rows in results table.
- Predict now has an A.I. bot.
- Lots of save model tidying.
- Added support for more than two description detail fields when saving model.
- Standalone mode gets clearer messaging around simulated overlay dismissal.

https://user-images.githubusercontent.com/2205926/119808038-93dd6a80-be98-11eb-8aa8-87d1624c5bf1.mov


Erin007:
I literally clapped when I watched the video in the PR description. SO much cleaner and crisper and nice attention to little details. 😍

src/UIComponents/Predict.jsx
```diff
@@ -109,10 +108,12 @@ class Predict extends Component {
...
         {this.props.predictedLabel && (
-          <div>
-            <p />
-            <div>A.I. predicts:</div>
-            <div style={styles.contents}>
+          <div style={{ ...styles.contents, marginTop: 20 }}>
+            <div style={{ float: "left", width: "20%" }}>
+              <img style={{ width: "100%" }} src={aiBotBorder} />
+            </div>
+            <div style={{ float: "right", width: "75%" }}>
```
Erin007, line 115: small nit: can we extract styles and put them in constants with the others for consistency?

src/UIComponents/SaveModel.jsx
```diff
@@ -59,38 +61,40 @@ class SaveModel extends Component {
...
 
     return fields;
   };
 
   render() {
-
     const nameField = {
       id: "name",
-      text: "What will you name the model? (required)"
+      text: "Model name:"
```
Erin007, line 90: Do we still want to indicate somehow that this is a required field? Or ok to just let users find that out if they hit an error by trying to save without a name?

src/constants.js
```diff
@@ -169,22 +178,32 @@ export const styles = {
...
   contents: {
     borderRadius: 0,
-    backgroundColor: "rgb(206, 206, 206)",
+    backgroundColor: "#f2f2f2",
     padding: 15
   },
 
+  contentsCsvButton: {
+    borderRadius: 0,
+    backgroundColor: "#f2f2f2",
```
Erin007, line 200: This color is used a few time - worth extracting into a variable? I think we have a `colors` object already we could use, but I don't feel strongly about it.

Erin007 (approved):

src/UIComponents/SaveModel.jsx
dancodedotorg: Ooo - good catch. I have a semi-strong opinion that we should keep the (required) part because of how it behaves with the "Save" button on this screen. AI Lab hides the "Save" button until you start typing in the model name (you can see this in the video around 1:57). This is the only time in the tool that this happens - all other times, the button is visible but disabled (you can see this with "Continue" and "Train" buttons earlier in the video). 

I worry that a user will get to that screen and see the Save button is missing, then think they did something wrong, when in reality they just need to start typing in Model Name section for it to appear (this has also happened to me a few times). I think including the "required" label would help cue a user that they need to do something here first, and then they would notice that the Save button appears (which has also happened to me).

(As I wrote this, I started to wonder: why does the Save button behave differently on this panel than the others?)

src/UIComponents/Predict.jsx
breville: Done.

src/UIComponents/SaveModel.jsx
breville: I restored `(required)`.  These concerns are valid.  For what it's worth, we do disable nav buttons in other screens until the user has done something required.

src/constants.js
breville: Done.

breville:
> * I think 'Extrema' for the min/max/range is too high on the vocabulary scale if our target audience is middle schoolers. I notice the label above the scatterplot is "Relationship information" - I wonder if we can use "Column information" for this section instead? That way, we can also use it for the bar charts with categorical data - both the bar chart and the min/max/range are column-specific info, and I think it makes it a little clearer what differentiates the 2 sections.

Done.

> * Just wondering - did the crosstab tables for categorical data get any style updates in this too? Or maybe that's waiting for the bar charts to appear in the column explorer panel ([like this card](https://trello.com/c/CUI4JdsA/210-show-bar-chart-distribution-in-feature-selection-screen)) to update those stylings together?

@dancodedotorg  I haven't done any CrossTab styling in this PR but can address it in a followup.  Are there specific changes we have in mind?

src/UIComponents/SaveModel.jsx
breville: I restored `(required)`.  These concerns are valid.  I can change the `Save` button be disabled, rather than hidden, in a followup.

dancodedotorg:
> @dancodedotorg I haven't done any CrossTab styling in this PR but can address it in a followup. Are there specific changes we have in mind?

I trust the improvements you and Mark are scoping; I was more just making sure I didn't miss anything when reviewing. I wouldn't want to say "This PR looks great!" only to go "actually, for the cross tabs..." later.

markabarrett:
I am going to add more feedback but wanted to start with a large, bold **WOW**

PR #198, opened 2021-05-27
Existing dataset cleanup

moneppo:
* Update images to vector versions
* Removes datasets that don't have interesting correlations that could make effective models
* Removed all rows with empty cells and some improvements to eliminate non-numeric data in numeric columns.


Erin007 (approved):

breville:
@moneppo We shouldn't have text in the images.  They will be impossible to localise and are duplicated by the name below the image in our UI.  Apologies for not catching earlier when the comp was shared; I thought that was for presentation purposes only.

Erin007 (commented):
Screen Shot 2021-05-27 at 4 01 17 PM

Just noticed that in addition to changing the dataset csvs we'll need to comb the JSON metadata to remove mention of deleted data. For example, the column description here references NA - which is fine for this case of categorical - but are there other numerical column descriptions that say something similar?

moneppo:
That’s a good call. I’ll review with Jael to make sure we have an updated
template too. I’ll yank these files from the PR in the interim.

moneppo:
No, I did a pass to make sure that any numerical column only contains
numerical data.

breville (approved):

PR #199, opened 2021-05-27
Hide selection button for columns with invalid data 

Erin007:
We got a [bug report](https://codedotorg.slack.com/archives/C016VGH1BT6/p1622129533102700?thread_ts=1622064539.067300&cid=C016VGH1BT6) that a student couldn't import their model into App Lab. The issue occurred because design elements and code couldn't be autogenerated. The error was `cannot read property toFixed of null`, which we hit because the max and min where set to null for one of the model's features (per capita crime rate).
Screen Shot 2021-05-27 at 4 43 13 PM

The max and min could not be calculated because there were invalid inputs in that numerical column (NA). Dataset work (e.g. #198) is underway to clean the in-house datasets which will prevent this problem from occurring. However, students may upload dirty data. This PR ensures that columns with invalid data (not numbers for numerical columns or categorical columns with too many unique values) can not be selected as features or labels. 

While making this change, I also hoisted some of the logic currently handled by the `ColumnInspector` component into more modular Redux functions (in the vein of #187). 

BEFORE: When clicking on a numerical column with dirty data, we showed the warning but still let students select the column. 
Screen Shot 2021-05-27 at 4 52 03 PM

AFTER: Users can click and inspect a column and see the warning, and do NOT have a select button. 
Screen Shot 2021-05-27 at 4 50 33 PM


Erin007:
cc: @dancodedotorg

src/redux.js
```diff
@@ -633,6 +633,45 @@ export function getSelectedColumns(state) {
...
+function isValidNumericalData(state, column) {
+  return !isNaN(getMaximumValue(state, column));
+}
+
+function isColumnDataValid(state, column) {
+  return (
+    isColumnCategorical(state, column) && isValidCategoricalData(state,column)
+  ) ||
+  (isColumnNumerical(state, column) && isValidNumericalData(state,column));
+}
```
Erin007, line 657: My intent is that this structure will allow us to layer in a check for empty cells, or anything else we find mucks up the data more easily, but definitely open to feedback!

breville: This seems fine for now, but if it gets much longer, I like the idea of a series of separate `if` statements and then a default return value if none of them hit.

src/UIComponents/ColumnInspector.jsx
```diff
@@ -79,17 +78,17 @@ class ColumnInspector extends Component {
...
-      currentColumnData &&
-      currentColumnData.dataType === ColumnTypes.CATEGORICAL
-    ) {
+    const isCategorical =
+      currentColumnData
+      && currentColumnData.dataType === ColumnTypes.CATEGORICAL;
+
+    const isNumerical =
+      currentColumnData
+      && currentColumnData.dataType === ColumnTypes.NUMERICAL;
```
breville, line 89: Could these conditions be done in redux too?

breville: (I don't have a strong opinion, though.)

breville (approved):
Good stuff!

PR #201, opened 2021-05-28
Show bar chart for categorical column in feature selection panel

breville:
### before

Screen Shot 2021-05-29 at 7 26 18 AM

### after

Screen Shot 2021-05-29 at 7 38 10 AM

### also after

Screen Shot 2021-05-29 at 7 39 17 AM

### DIFF
--- a/src/UIComponents/ColumnInspector.jsx
+++ b/src/UIComponents/ColumnInspector.jsx
@@ -9,7 +9,12 @@ import {
   addSelectedFeature,
   removeSelectedFeature
 } from "../redux";
-import { ColumnTypes, styles, colors, UNIQUE_OPTIONS_MAX } from "../constants.js";
+import {
+  ColumnTypes,
+  styles,
+  colors,
+  UNIQUE_OPTIONS_MAX
+} from "../constants.js";
 import { Bar } from "react-chartjs-2";
 import ScatterPlot from "./ScatterPlot";
 import CrossTab from "./CrossTab";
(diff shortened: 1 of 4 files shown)


Erin007:
Screenshots please.

breville:
@Erin007 Added screenshots.

Erin007 (approved):

dancodedotorg:
Thinking about this and the updates to Scatterplots in the other PR - I like that both sets of graphs use "Column Information" and "Relationship Information" to label those areas of this panel. I think that consistency is helpful for teachers when explaining the graphs to students.

Consistency question: In this screenshot, the column information (bar chart) is above the relationship information (cross tab). I _think_ for numerical data, these charts are switched - the relationship information (scatterplot) is above the column information (min/max/range). If possible, I think we should make it consistent between the two so it's easier to train students to look in the same place whether they're looking at numerical or categorical data.

Maybe we can move the crosstab above the bar chart in this view so it matches the scatterplot? I don't have strong feelings about the order though, I just think it should be the same for numerical and categorical views.

made-line:
Hooray for bar charts! Would we want to add consistent spacing between the different sections for numerical and categorical columns? Currently, there's more spacing in the column inspector for numerical data:
Screen Shot 2021-06-01 at 2 08 41 PM


Categorical: 
Screen Shot 2021-06-01 at 2 16 25 PM

breville:
@dancodedotorg Great catch.  I'm now putting Relationship Information before Column Information for both numerical and categorical comparisons.

@made-line Between some tweaks here, and merging in some other changes, the spacing should now be more consistent.

dancodedotorg (approved):

PR #203, opened 2021-05-28
Move max and min to placeholder

Erin007:
Updating how we show max and min for numerical feature inputs in the testing panel to be consistent with https://github.com/code-dot-org/code-dot-org/pull/40853. 

BEFORE 
Screen Shot 2021-05-28 at 3 17 09 PM

AFTER 
Screen Shot 2021-05-28 at 3 13 13 PM


breville (approved):
Great!

PR #204, opened 2021-05-28
Remove rows with empty cells 

Erin007:
Previously we were filtering out training data that was empty or undefined. This was problematic because we didn't remove the whole row if there was a "dirty" cell, we just removed that cell.  Which could result in mismatched lengths of `trainingExamples` and `trainingLabels` which could not be processed by the KNN algorithm, causing an error: 
```
index.js?9dc3:33 Uncaught TypeError: Cannot read property 'push' of undefined
    at new KNN (index.js?9dc3:33)
```
The intermediate fix was to remove empty cells from the csvs for our internally curated datasets (e.g. https://github.com/code-dot-org/ml-playground/pull/195). This change alleviates this problem programmatically and for student uploaded datasets by eliminating any full row containing any undefined or empty cells prior to attempting to derive training examples and labels. 

I added a small note indicating that we remove rows with empty cells to clarify for the observant student who finds themselves with a smaller dataset in the tool than they intended. 

No rows removed - no empty cell message:
Screen Shot 2021-06-01 at 10 07 20 AM

1 row removed:
Screen Shot 2021-06-01 at 10 04 22 AM

More than 1 row removed: 
Screen Shot 2021-06-01 at 10 06 11 AM

### DIFF
--- a/src/UIComponents/DataDisplay.jsx
+++ b/src/UIComponents/DataDisplay.jsx
@@ -25,7 +25,7 @@ class DataDisplay extends Component {
           <DataTable />
         </div>
         <div style={styles.footerText}>
-          There are {this.props.data.length} rows of data.
+          There are {this.props.data.length} rows of data. If rows contained empty cells, they were removed.
           {this.props.data.length > 100 &&
             <span>
               &nbsp;
--- a/src/UIComponents/DataDisplay.jsx
+++ b/src/UIComponents/DataDisplay.jsx
@@ -8,11 +8,16 @@ import { styles } from "../constants";
 
 class DataDisplay extends Component {
   static propTypes = {
-    data: PropTypes.array
+    data: PropTypes.array,
+    removedRowsCount: PropTypes.number
   };
 
   render() {
-    const { data } = this.props;
+    const { data, removedRowsCount } = this.props;
+    const removedRowsMsg =
+      removedRowsCount === 1
+      ? `${this.props.removedRowsCount} row was removed because it contained an empty cell.`
+      : `${this.props.removedRowsCount} rows were removed because they contained an empty cell.`;
 
     if (data.length === 0) {
       return null;
@@ -25,7 +30,13 @@ class DataDisplay extends Component {
           <DataTable />
         </div>
         <div style={styles.footerText}>
-          There are {this.props.data.length} rows of data. If rows contained empty cells, they were removed.
+          There are {this.props.data.length} rows of data.
+          {removedRowsCount > 0 &&
+            <span>
+              &nbsp;
+              {removedRowsMsg}
+            </span>
+          }
           {this.props.data.length > 100 &&
             <span>
               &nbsp;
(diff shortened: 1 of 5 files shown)


Erin007:
Thanks to @made-line's clear organized [notes](https://docs.google.com/document/d/1bGKrH5DNfjotzfgvCO3Z4af-mVLLbiBL-SY7gUOu7GU/edit) that made implementing this change straightforward!  🙌

Erin007:
FYI: @dancodedotorg this makes the tool more resilient to dirty data and removes the pressure on students to remove empty cells prior to training.  If you'd like to be more heavy handed in encouraging them to clean their data more thoroughly before training, let's chat because we can go in that direction too.

src/UIComponents/DataDisplay.jsx
```diff
@@ -25,7 +25,7 @@ class DataDisplay extends Component {
           <DataTable />
         </div>
         <div style={styles.footerText}>
-          There are {this.props.data.length} rows of data.
+          There are {this.props.data.length} rows of data. If rows contained empty cells, they were removed.
```
breville, line 28: I think we should say something like "Another X rows contained empty cells and were removed." and only show it if we actually removed rows.

breville: (The guiding principle here is that if the software knows the answer, in this case whether there were empty cells, then it can be specific about what it tells the user.)

Erin007: Updated code @ screenshots! Thanks for suggesting.

src/csvReaderWrapper.js
```diff
@@ -14,12 +18,37 @@ export const parseCSV = (csvfile, download, setColumnsToOther) => {
...
+  return cell !== undefined && cell !== "";
+}
+
+const isRowValid = (row) => {
+  var cells = Object.values(row);
+  for (var i = 0; i < cells.length; i++) {
+    if (!isCellValid(cells[i])) {
+      return false;
+    }
+  }
```
breville, line 31: I think we could use `cells.every` here?

src/UIComponents/DataDisplay.jsx
```diff
@@ -8,11 +8,16 @@ import { styles } from "../constants";
...
+    removedRowsCount: PropTypes.number
   };
 
   render() {
-    const { data } = this.props;
+    const { data, removedRowsCount } = this.props;
+    const removedRowsMsg =
+      removedRowsCount === 1
+      ? `${this.props.removedRowsCount} row was removed because it contained an empty cell.`
+      : `${this.props.removedRowsCount} rows were removed because they contained an empty cell.`;
```
made-line, line 20: Neat way to handle singular vs plural!

made-line (approved):

src/csvReaderWrapper.js
Erin007: Well, isn't that tidy? Thanks!

breville (approved):

PR #207, opened 2021-06-01
Autodetect column datatype for uploaded csvs, clean up variable name

Erin007:
`setColumnsToOther` is a relic of a time when we had three datatypes (numerical, categorical, other). Now, we have two: numerical and categorical so the variable name is confusing.  Instead, the intent of the `setColumnsToOther` boolean parameter is really something more like "should we infer the column datatype in the column from the data in the column because this is a user uploaded dataset that doesn't have column datatype set in JSON metadata?". I've updated to `useDefaultColumnDataType`. In addition to renaming the variable, we are now being more clever about setting the column's datatype. Previously, we would always default to categorical; now, if we detect that the column contains only numbers, we will set the column's datatype to numerical. 

https://user-images.githubusercontent.com/12300669/120349489-284a3180-c2cc-11eb-8b9d-aa79ee0e7b61.mov

### DIFF
--- a/src/csvReaderWrapper.js
+++ b/src/csvReaderWrapper.js
@@ -19,6 +19,10 @@ export const parseCSV = (csvfile, download, useDefaultColumnDataType) => {
   });
 };
 
+const isCellValid = (cell) => {
+  return cell !== undefined && cell !== "";
+}
+
 const isRowValid = (row) => {
   var cells = Object.values(row);
   return cells.every(isCellValid)
@@ -40,7 +44,7 @@ const updateData = (result, useDefaultColumnDataType) => {
   countRemovedRows(data, cleanedData);
   store.dispatch(setImportedData(cleanedData));
   if (useDefaultColumnDataType) {
-    setDefaultColumnDataType(data);
+    setDefaultColumnDataType(cleanedData);
   }
 }
 
--- a/src/csvReaderWrapper.js
+++ b/src/csvReaderWrapper.js
@@ -1,12 +1,16 @@
 import Papa from "papaparse";
 import { store } from "./index.js";
-import { setImportedData, setColumnsByDataType } from "./redux";
+import {
+  setImportedData,
+  setColumnsByDataType,
+  columnContainsOnlyNumbers
+} from "./redux";
 import { ColumnTypes } from "./constants.js";
 
-export const parseCSV = (csvfile, download, setColumnsToOther) => {
+export const parseCSV = (csvfile, download, useDefaultColumnDataType) => {
   Papa.parse(csvfile, {
     complete: result => {
-      updateData(result, setColumnsToOther);
+      updateData(result, useDefaultColumnDataType);
     },
     header: true,
     download: download,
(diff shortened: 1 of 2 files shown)


Erin007:
cc: @dancodedotorg this is a slight improvement for the scenario in which a user uploads a dataset. Instead of always defaulting to assume the columns contain categorical data, we now "guess" that a column is numerical if it contains only numbers. Students can still update datatype via the dropdown in the column inspection panel if our programatic guess is wrong.

src/redux.js
```diff
@@ -645,6 +645,15 @@ function isColumnNumerical(state, column) {
   return (state.columnsByDataType[column] === ColumnTypes.NUMERICAL);
 }
 
+export function columnContainsOnlyNumbers(data, column) {
+  for (let row of data) {
+    if (isNaN((row[column]))) {
+      return false;
+    }
+  }
```
breville, line 653: Can we use `row.every` here?

src/csvReaderWrapper.js
```diff
@@ -1,30 +1,38 @@
...
 const setDefaultColumnDataType = data => {
-  Object.keys(data[0]).map(column =>
-    store.dispatch(setColumnsByDataType(column, ColumnTypes.CATEGORICAL))
-  );
+  const columns = Object.keys(data[0]);
+  for (let column of columns) {
+    if (columnContainsOnlyNumbers(data, column)) {
+      store.dispatch(setColumnsByDataType(column, ColumnTypes.NUMERICAL))
+    } else {
+      store.dispatch(setColumnsByDataType(column, ColumnTypes.CATEGORICAL))
```
made-line, line 35: Would we want to handle a case where the column values are numbers but it's considered categorical (e.g., `Slope of Peak Exercise ST Segment` column in the Hearts dataset)?

Erin007: I thought about this, but wasn't convinced there was a fool-proof heuristic to auto-detect that case. What do you think? If the column contains only numbers and has less than X unique values? I opted not to do something like that because the more complexity we have behind the scenes the more we have to expose to students so they can reason about what's happening. We don't want to be _too_ clever with our guesses and we'll never be right 100% of the time, which is why we still allow students to change data type in the column inspector dropdown.

dancodedotorg: I think the line we're trying to walk is "helpful, not perfect". There's curricular value in having students think critically about data types, including situations where numbers actually represent categories (like # of legs in the Zoo dataset). I think doing our best to autodetect, but then giving students the option to change via the dropdown, is the best move here - it still gives students a lever of control and ownership over their data, without being so tedious that they need to update every single column (which was a problem in the past)

src/redux.js
Erin007: we sure can!

src/csvReaderWrapper.js
made-line: That makes sense! Thank you both for the thoughtful discussion.

made-line (approved):

breville (approved):

PR #208, opened 2021-06-01
Log panel navigation as a Google Analytics page view

Erin007:
To track how users move through the flow of AI Lab, I added Google Analytics logging that will send a metrics for each panel view as a page view by appending the current panel to the level url. For example: 

Screen Shot 2021-06-01 at 3 40 25 PM

The current panel is logged in coordination with setting `state.currentPanel`.

The typical panel flow is `selectDataset`, `dataDisplayLabel`, `dataDisplayFeatures`, `trainModel`, `generateResults`, `results`, `saveModel`,  then `modelSummary`. 

Screen Shot 2021-06-02 at 1 05 30 PM

### DIFF
--- a/src/redux.js
+++ b/src/redux.js
@@ -176,7 +176,6 @@ export function setInstructionsKeyCallback(instructionsKeyCallback) {
 }
 
 export function setCurrentPanel(currentPanel) {
-  reportPanelView(currentPanel);
   return { type: SET_CURRENT_PANEL, currentPanel };
 }
 
@@ -453,6 +452,7 @@ export default function rootReducer(state = initialState, action) {
     };
   }
   if (action.type === SET_CURRENT_PANEL) {
+    reportPanelView(action.currentPanel);
     let showedOverlay = false;
     if (state.instructionsKeyCallback) {
       const options = {};
--- a/src/redux.js
+++ b/src/redux.js
@@ -176,6 +176,7 @@ export function setInstructionsKeyCallback(instructionsKeyCallback) {
 }
 
 export function setCurrentPanel(currentPanel) {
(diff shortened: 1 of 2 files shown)


Erin007:
@breville Is there a way to check definitively that this works? I can pull up https://analytics.google.com/analytics/web/?authuser=1#/report/content-drilldown/a37745279w66217109p78966704/explorer-table.plotKeys=%5B%5D&_r.drilldown=analytics.pagePathLevel1:~2Fs~2F,analytics.pagePathLevel2:~2Fallthethings~2F,analytics.pagePathLevel3:~2Flessons~2F,analytics.pagePath:~2Fs~2Fallthethings~2Flessons~2F43~2Flevels~2F1/ but suspect we don't log from local environments.

breville:
@Erin007 The `Google Analytics Debugger` for Chrome appears to show what will happen, even for a local server.

breville:
We should record the page view when the panel is first set as the current panel, rather than when the user navigates away.  That way, we will get proper page view durations.

Erin007:
> We should record the page view when the panel is first set as the current panel, rather than when the user navigates away. That way, we will get proper page view durations.

Oh! Good call. I didn't think about page view durations. Let me re-work this.

src/redux.js
```diff
@@ -176,6 +176,7 @@ export function setInstructionsKeyCallback(instructionsKeyCallback) {
 }
 
 export function setCurrentPanel(currentPanel) {
+  reportPanelView(currentPanel);
```
breville, line 179: I think the idiomatic Redux approach would be to do the report in the `SET_CURRENT_PANEL` handler below?

breville (approved):

PR #209, opened 2021-06-02
Fixed Heart dataset bug

made-line:
This PR addresses a bug in the Heart dataset. When clicking on the Age column in the Heart dataset, the `ColumnInspector` component doesn't render because of an error:
`Uncaught TypeError: Cannot read property 'type' of undefined`

Screenshot:
Screen Shot 2021-06-01 at 11 53 07 PM

 
Before:

https://user-images.githubusercontent.com/40412372/120441601-96790d80-c339-11eb-8c99-0d51ecf62bb8.mov

After:

https://user-images.githubusercontent.com/40412372/120441669-a7298380-c339-11eb-92ae-999990be37f5.mov


Erin007:
> I wonder if we should also guard against this in code? We could at least output an informative error in the console if we detect this situation, rather than give a JS error.

In general I agree that we should include defensive programmatic fixes when possible, but I'm not sure it's worth it in this case. This will only ever be a problem for internal datasets since the root cause is a mismatch between CSV and JSON. @moneppo's dataset clean up work should take care of any problems we have in the current datasets. And I _think_ @dancodedotorg said that the the JSON-generator Brent set up should keep JSON and CSV column names consistent for any future datasets.

Erin007 (approved):

dancodedotorg (approved):

dancodedotorg:
>And I think @dancodedotorg said that the the JSON-generator Brent set up should keep JSON and CSV column names consistent for any future datasets.

FWIW: this is true - if we use Brent's metadata generator, it uses the CSV column name to generate the id so they'll always match

breville:
> In general I agree that we should include defensive programmatic fixes when possible, but I'm not sure it's worth it in this case. This will only ever be a problem for internal datasets since the root cause is a mismatch between CSV and JSON.

That is true, but I have seen this error quite a lot over the last few months, and it's not very self-evident, so we could do a little extra to make it easier to diagnose in the future, especially for future maintainers who are new to the code.

PR #212, opened 2021-06-03
Log accuracy each time a model is trained

Erin007:
AI Lab Analytics [notes](https://docs.google.com/document/d/1gf3hmksYL08vdU0xUPOHnSH_z5D0jqO19tpLIilwe48/edit)

We want to be able to determine if students are iterating while training models or not. The ideal scenario might look something like: 
Screen Shot 2021-06-03 at 4 29 02 PM
where students increase accuracy of their models by trying different feature combinations. By logging to Redshift each time a user trains a model, including label, features and accuracy we will be able to analyze if and when students iterate. 

Screen Shot 2021-06-03 at 4 32 03 PM

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -154,6 +154,7 @@ class App extends Component {
     validationMessages: PropTypes.object,
     onContinue: PropTypes.func,
     logMetric: PropTypes.func,
+    modelMetrics: PropTypes.object,
     resultsPhase: PropTypes.number,
     startSaveTrainedModel: PropTypes.func,
     dataToSave: PropTypes.object,
--- a/src/App.js
+++ b/src/App.js
@@ -45,8 +45,6 @@ class PanelButtons extends Component {
       this.props.onContinue();
     } else if (this.props.currentPanel === "saveModel") {
       this.props.startSaveTrainedModel(this.props.dataToSave);
-    } else if (this.props.currentPanel === "generateResults") {
-      this.props.setCurrentPanel(this.props.panelButtons.next.panel);
     } else {
       this.props.setCurrentPanel(this.props.panelButtons.next.panel);
     }
--- a/src/App.js
+++ b/src/App.js
@@ -18,8 +18,7 @@ import {
   validationMessages,
   getTrainedModelDataToSave,
   isSaveComplete,
-  shouldDisplaySaveStatus,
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -45,6 +48,9 @@ class PanelButtons extends Component {
       this.props.onContinue();
     } else if (this.props.currentPanel === "saveModel") {
       this.props.startSaveTrainedModel(this.props.dataToSave);
+    } else if (this.props.currentPanel === "generateResults") {
+      this.props.logMetric("train-model", this.props.modelMetrics);
```
breville, line 52: I think we should move the logging out of the UI and into `redux.js`..

Erin007: Hmmm. I played around with this and I'm not sure how/where else to put it. We want to log a metric every time a new model is trained, so we could store the logMetric function as a piece of state, import it into the trainer and log from `KNNTrainer` for example, but that somehow feels even messier? Keeping the logging in UI matches the pattern we have for the save function. Do you have other ideas?

breville: I don't have super strong feelings about this, but I was thinking we could put the log call into redux.js, specifically where rootReducer handles the SET_CURRENT_PANEL case.  That way, rather than UI component code doing logging, it's done at the point of a relevant state change.

But to your point, we can go even further and decouple from navigation UI entirely, and instead log when the model is trained.  We already dispatch to redux a couple times from KNNTrainer.js, with setTrainedModel and setModelSize, so we could move the log call into redux.js where it handles one of them?  

A benefit of either approach is that redux.js will no longer need to export getModelMetrics.  Similarly, we would not need to export logMetric any more either.  This is also similar to how we keep reportPanelView self-contained in redux.js.

The underlying principle here is that UI components trigger state changes, and respond to state changes, but carry out as little business logic as possible.

Erin007: 💡

Erin007: I'm a little worried about tying the logging to navigation because in this case there's a risk of overlogging. For example, if we log model X details when we get to the results screen, then go to the save screen, then click back, we'd be on results again and re-log model X details, which we don't want to do.  So, I think I'm leaning towards logging from KNNTrainer.

breville (approved):
Nice!

PR #213, opened 2021-06-04
Fix Windows Chrome CrossTab text

breville:
For some reason, the Windows version of Chrome doesn't support `writing-mode: vertical-rl;` on a `<td>`, while it works fine on macOS Chrome.  This is an [easy fix](https://stackoverflow.com/a/59854643).

### before

Screen Shot 2021-06-04 at 5 49 24 PM

### after

Screen Shot 2021-06-04 at 5 51 08 PM

### DIFF
--- a/src/UIComponents/CrossTab.jsx
+++ b/src/UIComponents/CrossTab.jsx
@@ -75,14 +75,15 @@ class CrossTab extends Component {
                   </thead>
                   <tbody>
                     <tr>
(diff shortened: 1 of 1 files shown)


Erin007 (approved):
Sneaky! Good find and fix.

made-line (approved):
Very interesting! Thanks for posting the link to the Stack Overflow post.

PR #214, opened 2021-06-04
ScrollableContent component

breville:
This change contains a small refactor to introduce a `ScrollableContent` component.

The change previously contained work to implement a vignette on scrollable content, to help users without scrollbars to  understand that there was more content.  We decided not to ship it for now, but could revisit if necessary using the work in this change's historical commits.  For posterity, here's how it was looking:

https://user-images.githubusercontent.com/2205926/121082746-65626800-c793-11eb-9f09-dec113134acc.mov

Note: much easier to read with [whitespace changes hidden](https://github.com/code-dot-org/ml-playground/pull/214/files?diff=unified&w=1).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@code-dot-org/ml-playground",
-  "version": "0.0.31",
+  "version": "0.0.32",
   "main": "dist/main.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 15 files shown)


made-line (approved):
Wow, what an awesome way to refactor something that's used in a lot of components! I didn't know you could make a component for styling like this.

PR #215, opened 2021-06-04
New vector-style dataset images

breville:
All datasets, at 1280px browser width:

ailab-dataset-images-1


made-line (approved):
LGTM!

Erin007:
Screen Shot 2021-06-07 at 11 19 54 AM

super nit: can we make naan line up with the others. Looks like the image is bigger?

dancodedotorg (approved):

breville:
> super nit: can we make naan line up with the others. Looks like the image is bigger?

@Erin007 Nice catch; now fixed.

PR #217, opened 2021-06-06
Fix table border leak

breville:
The sticky header used in tables was not entirely covering the scrolling content.  After a lot of reading and experimenting, this is apparently the solution to keep the majority of the existing appearance while avoiding the "leak".

### before

Screen Shot 2021-06-06 at 9 29 44 PM

### after

Screen Shot 2021-06-06 at 9 30 04 PM

### before

Screen Shot 2021-06-06 at 9 32 13 PM

### after

Screen Shot 2021-06-06 at 9 33 43 PM

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -313,9 +313,9 @@ export const styles = {
 
   displayTable: {
     whiteSpace: "nowrap",
-    borderSpacing: 0,
+    borderSpacing: 0.5,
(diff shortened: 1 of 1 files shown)


Erin007 (approved):
Thank you! I noticed this, but forgot about it in the shuffle of everything. Thanks for tending to the details.

PR #218, opened 2021-06-07
Save status for pii and profanity 

Erin007:
https://github.com/code-dot-org/code-dot-org/pull/40915

Prior to saving we are now passing trained model data through a profanity and personally identifying information filter. If something unsafe to share is caught, we block the save of the model and show this message in AI Lab: 

Screen Shot 2021-06-07 at 12 17 20 PM

### DIFF
--- a/src/UIComponents/SaveModel.jsx
+++ b/src/UIComponents/SaveModel.jsx
@@ -18,7 +18,6 @@ class SaveModel extends Component {
     trainedModelDetails: PropTypes.object,
     labelColumn: PropTypes.string,
     columnDescriptions: PropTypes.array,
-    saveStatus: PropTypes.string,
     dataDescription: PropTypes.string,
     isUserUploadedDataset: PropTypes.bool
   };
@@ -229,7 +228,6 @@ export default connect(
     trainedModelDetails: state.trainedModelDetails,
     labelColumn: state.labelColumn,
     columnDescriptions: getSelectedColumnDescriptions(state),
-    saveStatus: state.saveStatus,
     dataDescription: getDataDescription(state),
     isUserUploadedDataset: isUserUploadedDataset(state)
   }),
--- a/src/constants.js
+++ b/src/constants.js
@@ -32,7 +32,7 @@ export const UNIQUE_OPTIONS_MAX = 50;
 export const saveMessages = {
   success: "Your model was saved!",
   failure: "There was an error. Your model did not save. Please try again.",
-  piiProfanity: "Your model could not be saved because it contains profanity or personally identifying information (e.g. email, address, phone number, etc). "
+  piiProfanity: "Your model could not be saved because it contains profanity or personally identifying information (e.g. email, address, phone number, etc)."
 };
 
 export function getFadeOpacity(animationProgress) {
--- a/src/constants.js
+++ b/src/constants.js
@@ -31,7 +31,8 @@ export const UNIQUE_OPTIONS_MAX = 50;
 
 export const saveMessages = {
   success: "Your model was saved!",
-  failure: "There was an error. Your model did not save. Please try again."
+  failure: "There was an error. Your model did not save. Please try again.",
+  piiProfanity: "Your model could not be saved because it contains profanity or personally identifying information (e.g. email, address, phone number, etc). "
 };
 
 export function getFadeOpacity(animationProgress) {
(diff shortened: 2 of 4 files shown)


src/UIComponents/SaveModel.jsx
```diff
@@ -229,7 +228,6 @@ export default connect(
     trainedModelDetails: state.trainedModelDetails,
     labelColumn: state.labelColumn,
     columnDescriptions: getSelectedColumnDescriptions(state),
-    saveStatus: state.saveStatus,
```
Erin007, line 232: cleaning up... this prop wasn't being used in this component, so I deleted it.

src/constants.js
```diff
@@ -31,7 +31,8 @@ export const UNIQUE_OPTIONS_MAX = 50;
 
 export const saveMessages = {
   success: "Your model was saved!",
-  failure: "There was an error. Your model did not save. Please try again."
+  failure: "There was an error. Your model did not save. Please try again.",
+  piiProfanity: "Your model could not be saved because it contains profanity or personally identifying information (e.g. email, address, phone number, etc)."
```
breville, line 35: nit: Apparently there is no need to have "etc." when starting with "e.g.", because it's a list of examples, rather than being exhaustive.

src/redux.js
```diff
@@ -255,7 +255,7 @@ const initialState = {
   currentPanel: "selectDataset",
   currentColumn: undefined,
   resultsPhase: undefined,
-  // Possible values for saveStatus: notStarted, started, success, and failure.
+  // Possible values for saveStatus: notStarted, started, success, piiProfanity, and failure.
```
breville, line 258: nit: these are all strings, so they could be in quotes?

src/UIComponents/SaveModel.jsx
made-line: Thank you for catching this!

made-line:
I'm unable to get consistent results when testing the profanity filter. Sometimes I'm able to save a model with "naughty" in the title:
Screen Shot 2021-06-07 at 5 20 43 PM

When I'm not able to save, I get the message: `There was an error. Your model did not save. Please try again.` Does this happen on your end?
Screen Shot 2021-06-07 at 5 25 59 PM

Erin007:
@made-line 
 If you want to test this out, there are some additional steps needed to get filtering to work locally. This line: https://github.com/code-dot-org/code-dot-org/blob/8d4a1e235c8867ff6a831d2db77c0713eab4b505/lib/cdo/web_purify.rb#L18 prevents actually hitting the webpurify API from local environments.  To test this, I comment out that line, then add in a webpurify key on this line: https://github.com/code-dot-org/code-dot-org/blob/8d4a1e235c8867ff6a831d2db77c0713eab4b505/lib/cdo/web_purify.rb#L26. That key lives in the Secrets Manager in AWS. When that's all in place, then you can reasonably expect the filter to actually be checking for potential profanities. "Naughty" isn't a scandalous enough word to actually trigger the filter :) You'll have to use an actual swear word (or email, phone number, etc). This explains why you're never hitting the new piiProfanity message case, but doesn't explain why you're hitting the generic failure case sometimes. Are you seeing any console errors when the save doesn't succeed and the filter is off?

breville (approved):
Nice!

made-line:
Thanks for the info! I followed those steps. When I try to save a model with an email address, I get the profanity message. When I try to save a model with a curse word or any non-profanity word, I get the `There was an error. Your model did not save. Please try again.` message. This is the console error I see: 
Screen Shot 2021-06-09 at 12 04 17 PM

I also tested this on Levelbuilder and I get the same error when I try to save a model with a curse word or non-profanity word. When I use an email address, I see the loading spinner and then no error message. I'm not able to navigate forward.

Erin007:
> Thanks for the info! I followed those steps. When I try to save a model with an email address, I get the profanity message. When I try to save a model with a curse word or any non-profanity word, I get the `There was an error. Your model did not save. Please try again.` message. This is the console error I see:
> Screen Shot 2021-06-09 at 12 04 17 PM
> 
> I also tested this on Levelbuilder and I get the same error when I try to save a model with a curse word or non-profanity word. When I use an email address, I see the loading spinner and then no error message. I'm not able to navigate forward.

Hmmm. I don't repo, but maybe you've stumbled on something I'm missing. Let's debug together tomorrow?

PR #220, opened 2021-06-07
Results screen headers get an Accuracy caption

breville:
Screen Shot 2021-06-08 at 9 26 58 AM

Also, upon return to the Results screen, a previously-shown ResultsDetails popup no longer shows.


src/constants.js
```diff
@@ -490,6 +490,10 @@ export const styles = {
     width: "75%"
   },
 
+  resultsHeader: { width: "75%", float: "left" },
+
+  resultsHeaderAccuracy: { fontSize: 14, position: "relative", top: 16 },
```
made-line, line 495: nit: do we want to format this on multiple lines for consistency?

breville: This is how JsPrettier did the formatting, and occurs elsewhere in this file too, so I think it's okay.

made-line (approved):

PR #221, opened 2021-06-08
Layout datasets before images load

breville:
This change uses the assumption that dataset images have a 1.5 aspect ratio (e.g. 720x480) to layout the SelectDataset scene with correct-sized tiles before the actual image files are loaded. Previously, it depended upon the images being loaded to get the correct layout, which made for a bit of a mess during the image loading.

(This is a fresh attempt after https://github.com/code-dot-org/ml-playground/pull/216 was merged prematurely into a working branch.)


made-line (approved):
LGTM!

PR #223, opened 2021-06-09
Navigation validation extraction

Erin007:
This is a clean-up PR with two main themes: 1.) removing unused validation messages and 2.) extracting navigation validation logic into its own file. 

We were no longer using the `validationMessages` function in redux.js and most of the logic covered in it was parceled out to other updates. For example, `ColumnInspector` now renders an error message if a column with strings is set as numerical (#199) and the empty cell validation code is moot given #204. I also removed the pointless references to `validationMessages` in Apps.js. 

While doing this, it became clear that most of the code in validate.js was no longer needed, or if it was needed, was actually conceptually related to whether panels were available based on whether certain pieces of state were set and their values. 

Thus, motivated by a recent Dev Book Club discussion about reducing mega long files by extracting related ideas into more specifically named files, I re-branded `validate.js` into `navigationValidation.js` and I moved over ~200 lines of panel button related logic. redux.js now only has 2 imports from `navigationValidation.js` and `navigationValidation.js` doesn't import anything from anywhere.


breville (approved):
This is great refactoring!

src/navigationValidation.js
```diff
@@ -0,0 +1,220 @@
...
+  { id: "modelSummary", label: "Finish" }
+];
+*/
+
+// Is a panel ready to be visited?  This determines whether a visible
+// nav button is enabled or disabled.
+export function isPanelEnabled(state, panelId) {
+
+  if (panelId === "dataDisplayLabel") {
+    if (!isDataUploaded(state)) {
```
maureensturgeon, line 27: Alternatively, you could also replace the inner `if` with `return isDataUploaded(state)`

maureensturgeon (commented):
Great clean up, thanks for sharing!

PR #224, opened 2021-06-09
Redux refactor

Erin007:
This is a sizable refactor of redux.js including a few side clean-up tasks along the way: 

- Removed unused functions (`getFeatures`, `getSelectedNumericalColumns`, `getSelectableFeatures`, `getSelectableLabels`) and stopped exporting functions that aren't used outside of redux.js

- extracted `getModelMetrics`, `logFirehoseMetric` and `reportPanelView` into a metrics-specific helper file

- extracted functions related to getting details about a specific column into a `columnDetails.js` helper file 

- changed `summaryStat` to `percentCorrect` when passing to `ModelCard` and `ResultsDetails` so we don't pass more information than is needed and so we can avoid dipping into nested state in the UI 

- extracted functions related to calculating accuracy and grades into the a helper file `accuracy.js` and moved tests for `getAccuracyRegression` and `getAccuracyClassification` into the complementary test file 

- extracted generic util functions into a `utils.js` helper file 

- extracted functions for converting data into Machine Learning friendly integers and back into human-readable strings into a `valueConversions.js` file 

- Reorganized the remaining functions in `redux.js` so they are grouped roughly by purpose and ordered roughly by use in the training process and included comments for each grouping 

- Set up placeholder test files for newly created helper files to facilitate writing upcoming tests


src/accuracy.js
```diff
@@ -0,0 +1,78 @@
+import { ResultsGrades, REGRESSION_ERROR_TOLERANCE, } from "./constants.js";
```
breville, line 1: It would be great to have a comment at the top of each file explaining its general purpose.

src/accuracy.js
```diff
@@ -0,0 +1,78 @@
...
+import { getExtrema } from "./columnDetails.js";
+import { getConvertedAccuracyCheckExamples, getConvertedLabels, isRegression } from "./redux.js"
+
+export function getAccuracyClassification(state) {
```
breville, line 5: It would also be great to have a comment per function explaining its own purpose & usage.

breville:
This is looking great.  One idea is to move some of the more utilitarian files into a directory alongside `UIComponents/` so that `src/` contains a smaller set of more foundational code.  I think that will make it easier to find the entrypoints to the code when browsing it for the first time, amongst other benefits.

src/accuracy.js
Erin007: This feels excessive to me. I think the context and naming of the function and its parameters should sufficiently document the purpose. If there is a function that could be refactored for clarity - happy to revise!

src/helpers/accuracy.js
```diff
@@ -0,0 +1,88 @@
+/*
+  Functions for calculating the accuracy of trained machine learning models.
+  "Grade" is "correct" or "incorrect". Categorical predicted and expected labels must match exactly to count as "correct". Numerical predicted and expected labels must be within 5% of the range of labels to count as
```
breville, line 3: nit: very long line; should wrap.

src/helpers/utils.js
```diff
@@ -0,0 +1,22 @@
+/* Generic helper functions. */
+
+export function getKeyByValue(object, value) {
```
maureensturgeon, line 3: We could probably use some methods from lodash instead of defining our own for these helpers. Here we might use https://lodash.com/docs/4.17.15#findKey

src/helpers/utils.js
```diff
@@ -0,0 +1,22 @@
...
+  return Object.keys(object).find(key => object[key] === value);
+}
+
+export function isEmpty(object) {
```
maureensturgeon, line 7: Could use https://lodash.com/docs/4.17.15#isEmpty

src/helpers/utils.js
```diff
@@ -0,0 +1,22 @@
...
+  return Object.keys(object).length === 0;
+}
+
+export function areArraysEqual(array1, array2) {
```
maureensturgeon, line 11: Could use https://lodash.com/docs/4.17.15#isEqual

src/helpers/utils.js
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    array1.length === array2.length &&
+    array1.every((value, index) => {
+      return value === array2[index];
+    })
+  );
+}
+
+export function getRandomInt(max) {
```
maureensturgeon, line 20: Could use https://lodash.com/docs/4.17.15#random

test/unit/utils.test.js
```diff
@@ -0,0 +1,5 @@
+describe("example test", () => {
```
maureensturgeon, line 1: Did you mean to leave these empty tests?

test/unit/valueConversion.test.js
```diff
@@ -0,0 +1,5 @@
+describe("example test", () => {
```
maureensturgeon, line 1: Another empty test!

src/helpers/utils.js
Erin007: @breville I vaguely remember early in the project you had some ideas about not using libraries like lodash? Do you still feel that way?

test/unit/utils.test.js
Erin007: Yes! Placeholders for upcoming work to add tests. Arguably should've waited to include those files for when I start those tests (my next task), but don't see harm in having them here until then 🤷‍♀️

src/helpers/utils.js
breville: That sounds quite possible, but I don't have strong feelings now :)

breville (approved):

src/redux.js
```diff
@@ -1259,20 +947,77 @@ export function getIncorrectResults(state) {
...
+  return getColumnDataToSave(state, state.labelColumn);
+}
+
+function getSummaryStat(state) {
+  let summaryStat = {};
+  summaryStat.type = isRegression(state)
+    ? MLTypes.REGRESSION : MLTypes.CLASSIFICATION;
+  summaryStat.stat = getPercentCorrect(state);
+  return summaryStat;
+}
```
made-line, line 1002: Do we still need `getSummaryStat` if we're using `getPercentCorrect`?

src/redux.js
```diff
@@ -739,65 +694,38 @@ export function getSelectedCategoricalFeatures(state) {
...
+      return { id: columnId, readOnly: isColumnReadOnly(state, columnId) };
+    });
 }
 
-export function getUniqueOptions(state, column) {
-  return Array.from(new Set(state.data.map(row => row[column]))).filter(
-    option => option !== undefined && option !== ""
-  );
-}
+/* Functions for getting specific details about a batch of columns.  */
```
made-line, line 720: Could the functions below this comment go into the `columnDetails.js` file?

src/redux.js
```diff
@@ -810,295 +738,34 @@ export function getUniqueOptionsByColumn(state) {
...
-  });
-}
+/* Function for retriving aggreate details about a currently selected column. */
 
-export function getColumnDescription(state, columnId) {
-  if (!state || !columnId) {
+export function getCurrentColumnData(state) {
+  if (!state.currentColumn) {
     return null;
   }
```
made-line, line 752: Could this function also go into the `columnDetails.js` file?

made-line (approved):
Awesome work! It's nice to see how functions can be refactored and organized.

src/redux.js
Erin007: yes, because we still use in the `getDataToSave` function because we want to save the type of machine learning (regression or classification) alongside the percentage.

src/redux.js
Erin007: The distinction I made between the functions in `columnDetails` and the functions in this section are that `columnDetails` functions return information about a single specified column, whereas those in this section of redux.js are about multiple columns as a collection. In general, I kept anything that is being called directly from a UI component in the redux.js file so that it was accessible via connect.

made-line: Makes sense, thank you!

PR #226, opened 2021-06-10
move row removal message to data upload panel

Erin007:


src/UIComponents/DataCard.jsx
```diff
@@ -24,6 +25,11 @@ class DataCard extends Component {
       );
     }
 
+    const removedRowsMsg =
+      removedRowsCount === 1
+      ? `${this.props.removedRowsCount} row was removed because it contained an empty cell.`
+      : `${this.props.removedRowsCount} rows were removed because they contained an empty cell.`;
```
breville, line 31: suggest: " rows were removed because they contained empty cells."

breville, line 31: It seems slightly odd that we still generate this string for 0 removed rows, even though we don't show it later.

made-line (approved):
Tiny nit: would we want to add an asterisk to the number of rows and to the `removedRowsMsg` to visually indicate we made changes to the user's uploaded file? 
Screen Shot 2021-06-14 at 3 56 41 PM

src/UIComponents/DataCard.jsx
```diff
@@ -24,6 +25,13 @@ class DataCard extends Component {
       );
     }
 
+    const removedRowsMsg =
+      removedRowsCount > 0
+        ? removedRowsCount === 1
+          ? `${this.props.removedRowsCount} row was removed because it contained an empty cell.`
+          : `${this.props.removedRowsCount} rows were removed because they contained empty cells.`
+        : null;
+
```
breville, line 34: Possible other way to write this:

```
    const removedRowsMsg =
      removedRowsCount === 1
        ? `${this.props.removedRowsCount} row was removed because it contained an empty cell.`
        : removedRowsCount > 1
        ? `${this.props.removedRowsCount} rows were removed because they contained empty cells.`
        : null;

breville (approved):

PR #227, opened 2021-06-10
Changing normalized values back to actual temperature values.

moneppo:
Using the JS code `temp = (47 * temp - 8).toFixed(2)` and `atemp= (66 * atemp - 16).toFixed(2)`, I converted the normalized values in these columns to the original celsius values. These formulas were based on the info in the dataset source (https://archive.ics.uci.edu/ml/datasets/Bike+Sharing+Dataset):

- temp : Normalized temperature in Celsius. The values are derived via (t-t_min)/(t_max-t_min), t_min=-8, t_max=+39 (only in hourly scale)
- atemp: Normalized feeling temperature in Celsius. The values are derived via (t-t_min)/(t_max-t_min), t_min=-16, t_max=+50 (only in hourly scale)


moneppo:
I just verified that since the minimums listed are negative, my equations are correct. I wonder if they miswrote the minimums to actually be positive? Doesn't seem likely considering the standard range of temperatures in DC: https://www.currentresults.com/Weather/US/washington-dc-temperatures-by-month-average.php

The equations they list also make sense considering the kind of conversion being done.

breville:
@moneppo You are totally right.  I didn't see the minus signs at all, and my comment was wrong.  Thanks for verifying!

breville (approved):

PR #228, opened 2021-06-12
Unit tests for core ML functionality

breville:
Two unit tests for the core ML functionality: one for numerical data, and one for categorical.


test/unit/train.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+    train.init(store);
+    train.onClickTrain(store);
+
+    store.dispatch(setTestData({temperature: "150", rain: "1030"}));
+
+    train.onClickPredict(store);
+
+    const predictedValue = getConvertedPredictedLabel(store.getState());
+    expect(predictedValue).toBe(30);
```
Erin007, line 43: Hmmm. Looking at the training data, I'd expect the predictedValue to be 35 since the most similar training data point where temp is 150 and rain is 1030 is 35, not 30.  However, I do think it's valuable to have coverage while we figure out how to handle this situation, and getting this test set up merged in will allow us to iterate more quickly. Could we maybe include a comment about how if the K value optimization goes sideways with really small datasets such as this sample one?

test/unit/train.test.js
```diff
@@ -0,0 +1,79 @@
...
+    train.init(store);
+    train.onClickTrain(store);
+
+    store.dispatch(setTestData({flavor: "sweet", texture: "crunchy"}));
+
+    train.onClickPredict(store);
+
+    const predictedLabel = getConvertedPredictedLabel(store.getState());
+
+    expect(predictedLabel).toBe("yellow");
```
Erin007, line 77: similar to above, given what we know training is currently doing, "yellow" makes sense but we'd like to tweak things so that the expected is actually "green".

Erin007 (approved):
I'm ok with this going in as is, but would as we muddle through the accuracy wonkiness I'd like to expand the test suite to include more variety - large datasets, small datasets, datasets with all unique feature values and labels, datasets with duplicates etc There's some complexity buried in here that I don't want us to gloss over.

test/unit/train.test.js
Erin007: Oh, actually this makes sense, because the model is never trained on ` { temperature: "150", cost: "35", rain: "1030" }` because we pull that row to use as test data to calculate accuracy.

Erin007: Is there a way to set up these tests so we pass the training data rather than the whole dataset?

Erin007:
@breville how would you feel about merging #234 into this branch?

src/train.js
```diff
@@ -74,16 +72,18 @@ const extractTrainingLabel = (state, row) => {
...
+    extractTrainingExamples(updatedState, row)
+  );
+  const trainingLabels = updatedState.data.map(row =>
+    extractTrainingLabel(updatedState, row)
+  );
   /*
   Select X% of examples and corresponding labels from the training set to use for a post-training accuracy calculation.
   */
-  const numToReserve = parseInt(
+  const numToReserve = Math.ceil(
```
breville, line 86: @Erin007 Actually, will we always want to round up?

test/unit/train.test.js
breville: Good question.  This is something of an architecture question, and comes down to boundaries between modules.  I chose to unit test the exports from `train.js`.  We could unit test at a deeper level, or break apart some of the steps encapsulated in `init`, but I think we should do that refactoring separately.  This is a good example of how API designs can be informed by unit testing, and vice versa.

src/train.js
Erin007: I can't think of a time when we wouldn't want to round up.

PR #233, opened 2021-06-15
Improvements to TrainModel & GenerateResults animations

breville:
`TrainModel` & `GenerateResults` now show the full animation before beginning a fade out, in situations when we show all rows, rather than a subset, because there aren't many rows to train on or to test with.

`GenerateResults` now shows the actual predictions, rather than faking it with content from the source data set.  Because we use the `DataTable` component to show the data on this screen, I cheated a little and wrote a function that converts from the results format back into the input format for the `DataTable` to minimise code churn.

Also fixed the layout on these two screens to always show the heading at the top, even for a small amount of data.

https://user-images.githubusercontent.com/2205926/122514665-22896700-cfc1-11eb-92ba-1fa6a7c0f901.mov


src/redux.js
```diff
@@ -973,6 +981,35 @@ export function getIncorrectResults(state) {
...
+// Return results data so that it looks like regular data provided to the
+// DataTable.
+export function getResultsDataInDataTableForm(state) {
+  const resultsByGrades = getAllResults(state);
+
+  if (!resultsByGrades || resultsByGrades.examples.length === 0) {
+    return null;
+  }
+
+  const numItems = Math.min(10, resultsByGrades.examples.length);
```
made-line, line 997: Out of curiosity, why is 10 passed in here but 7 is passed in to the `getNumItems` function in `GenerateResults`?

breville: Great question.  I picked this arbitrarily, since none of the existing users of the code need more than 10, though you're right that they specifically don't need more than 7 at the moment.  I think I'll leave this value as-is, but add a comment explaining why.

src/redux.js
```diff
@@ -973,6 +981,35 @@ export function getIncorrectResults(state) {
   return getResultsByGrade(state, ResultsGrades.INCORRECT);
 }
 
+export function getAllResults(state) {
+  return getResultsByGrade(state, ResultsGrades.ALL);
+}
```
made-line, line 986: Is this function different from `getAccuracyGrades` in the `accuracy.js` file?

made-line (approved):
Looks awesome!

src/redux.js
breville: I believe so, because `getResultsByGrade` does conversion steps before filtering (via calls to `getConvertedAccuracyCheckExamples` and `getConvertedLabels`).

Erin007 (approved):

PR #234, opened 2021-06-15
Test ML 2.0 

Erin007:
Riffing off of #228 I made some adjustments to K values for certain cases - small datasets and classification v regression - to have passing tests for expected values that are more aligned to what we'd actually expect. For classification, when the dataset is small we use a K value that is the entire training dataset size so as to avoid hard-to-reason-about tie breaker situations. For regression, when the dataset is small, just look at the single closest neighbor. 

This is continued iteration on K value optimization described in #166.

### DIFF
--- a/src/trainers/KNNTrainer.js
+++ b/src/trainers/KNNTrainer.js
@@ -30,14 +30,27 @@ export default class KNNTrainer {
     this.store = store;
     const state = store.getState();
     const datasetSize = state.data.length;
-    const defaultRegressionK = datasetSize < 100 ? 1 : 5;
+    /*
+      We modify algorithm hyperparameters (k) based on dataset size and type of
+      machine learning in attempt to increase the liklihood of accurate
+      models that behave in ways consistent with the mental model presented in
+      the curriculum.
+    */
+    const smallDatasetSize = 10;
+    const mediumDatasetSize = 100;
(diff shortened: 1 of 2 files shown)


Erin007:
@dancodedotorg as an FYI this will change model predictions and accuracy for small models (I think in a way we want). I wrote up some more detailed [notes](https://docs.google.com/document/d/1nXbL-1iHYWyzVGpLB1aXzo2fptbRscD0A7ack1zl3zY/edit?usp=sharing).

src/trainers/KNNTrainer.js
```diff
@@ -29,34 +29,51 @@ export default class KNNTrainer {
   startTraining(store) {
     this.store = store;
     const state = store.getState();
+    const datasetSize = state.data.length;
+    const defaultRegressionK = datasetSize < 100 ? 1 : 5;
```
breville, line 33: It would be good to have some summary commentary capturing why `100`, `5`,  and `11` below, which all appear to be chosen deliberately...

breville (approved):

src/trainers/KNNTrainer.js
Erin007: Good idea. No magic numbers! Updated ✔️

PR #235, opened 2021-06-15
Tests for accuracy helper functions 

Erin007:


test/unit/accuracy.test.js
```diff
@@ -77,15 +101,44 @@ const resultsStateClassification = {
...
+    },
+    'play': {
+      'yes' : 0,
+      'no' : 1
+    },
+    'weather': {
+      'sunny' : 0,
+      'overcast' : 1,
+      'rainy': 2
+    }
```
Erin007, line 120: This set up feels particularly fragile because the `accuracyCheckLabels` and `accuracyCheckExamples` have to match up with the values in the `featureNumberKey` so the test data literally reads like 1s and 0s which is hard to puzzle through unless you know the system really well. Anyone have ideas for improvements?

test/unit/accuracy.test.js
```diff
@@ -32,12 +38,30 @@ const resultsState = {
...
+  accuracyCheckExamples: [[2], [2], [1], [1], [0], [0]],
+  selectedFeatures: ['sun'],
+  columnsByDataType: columnsByDataType,
+  featureNumberKey: {
+    'sun': {
+      'low' : 0,
+      'medium' : 1,
+      'high' : 2,
+    }
+  }
```
Erin007, line 52: There's a lot of state to set for these tests, which I suspect will be reusable in other upcoming tests. How do folks feel if I extract test data into its own file to re-use?

test/unit/accuracy.test.js
breville: I think I noticed something similar while working on [this](https://github.com/code-dot-org/ml-playground/pull/233/files#diff-7f797aa32260ac4cbc8b43b04d62976c18e503bea6cc86be7448f935896ec3cfR1000-R1021) code.  It's tricky when state involves parallel arrays that aren't explicitly tied to each other.  I think one solution is in a simpler representation in the core state.  Would it be possible to combine these parallel arrays into an array of objects?

test/unit/accuracy.test.js
breville: No objections to factoring as feels appropriate.

breville (approved):

test/unit/accuracy.test.js
Erin007: Ok, I think I'm going to get these tests in and consider a refactor separately.

PR #237, opened 2021-06-16
Clean up: Refactor KNNTrainer 

Erin007:
After the changes in #234 there were cases - small datasets -  where `predictedLabels` wasn't being set. In revisiting the code to fix that omission, I realized that the previous `startTraining` function was repetitive; each different case was tasked with doing the work of setting `this.knn`, `bestModel`, `bestK` and `bestPredictedLabels`. In this PR I extracted that shared work into a more modular and specifically named `getOptimalModelDetails`. 

Now the high level function `startTraining` handles calling helper functions to 1.) get the best model and its details 2.) store those model details in state, 3.) log a metric that a model was trained and 4.) store the historic results in state. 

`getOptimalModelDetails` determines which model is best by iterating through a set of possibleKValues, which in turn are determined by dataset size and type of machine learning (regression or classification).

### DIFF
--- a/src/trainers/KNNTrainer.js
+++ b/src/trainers/KNNTrainer.js
@@ -27,10 +27,7 @@ export default class KNNTrainer {
 
     logFirehoseMetric("train-model", state2);
 
-    const accuracy = getPercentCorrect(state2);
-    store.dispatch(
-      setHistoricResult(state2.labelColumn, state2.selectedFeatures, accuracy)
-    );
+    this.storeHistoricResult(store, state2);
   }
 
   /*
@@ -100,6 +97,17 @@ export default class KNNTrainer {
     return kValues;
   }
 
+  calculatePotentialKValues(state) {
+    let datasetSize = state.data.length;
+    let trainingExamplesSize = state.trainingExamples.length;
+    let possibleKValues = [1, 3, 5, 7, 17, 31, 45, 61];
+    const heuristicK = Math.round(Math.sqrt(datasetSize));
+    possibleKValues.push(heuristicK);
+    const oneThird = Math.round(datasetSize / 3);
+    possibleKValues.push(oneThird);
+    return possibleKValues.filter(kValue => kValue <= trainingExamplesSize);
+  }
+
   getAccuracyPercent() {
     const state = this.store.getState();
     const percent = getPercentCorrect(state);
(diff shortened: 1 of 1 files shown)


Erin007:
In this book-club-inspired refactor I tried to both modularize the code by extracting smaller functions and leverage better vertical formatting by arranging functions with increasing detail as we move down the file. What do you think @maureensturgeon @cforkish @davidsbailey?

src/trainers/KNNTrainer.js
```diff
@@ -14,96 +14,98 @@ import { logFirehoseMetric } from "../helpers/metrics";
...
+  */
+  getOptimalModelDetails(state) {
+    let optimalValues = {};
+    let bestModel = null;
+    let bestPredictedLabels = [];
+    let bestK = -1;
+    let bestAccuracy = -1;
+    const kValues = this.possibleKValues(state);
+    kValues.forEach(kValue => {
+      this.knn = new KNN(state.trainingExamples, state.trainingLabels, {
```
maureensturgeon, line 48: I'm curious, is there a reason we need to assign the result to `this.knn` and then assign that to `var model`? Or could we just assign the result directly to `model`?

src/trainers/KNNTrainer.js
```diff
@@ -14,96 +14,98 @@ import { logFirehoseMetric } from "../helpers/metrics";
...
+        bestAccuracy = accuracy;
+        bestK = kValue;
+        bestModel = model;
+        bestPredictedLabels = predictedLabels;
+      }
+    });
+    optimalValues.model = bestModel;
+    optimalValues.predictedLabels = bestPredictedLabels;
+    optimalValues.kValue = bestK;
+    return optimalValues;
```
maureensturgeon, line 66: Another option here would be to not assign a variable `optimalValues`, and just return 
```
{
  model: bestModel,
  predictedLabels: bestPredictedLabels,
  kValue: bestK
}
```

src/trainers/KNNTrainer.js
```diff
@@ -14,96 +14,98 @@ import { logFirehoseMetric } from "../helpers/metrics";
...
+  }
+
+  possibleKValues(state) {
```
maureensturgeon, line 69: This is sooo much easier to read!!

src/trainers/KNNTrainer.js
```diff
@@ -14,96 +14,98 @@ import { logFirehoseMetric } from "../helpers/metrics";
...
-    logFirehoseMetric("train-model", state2);
+    return kValues;
+  }
 
-    const accuracy = getPercentCorrect(state2);
-    store.dispatch(
-      setHistoricResult(state2.labelColumn, state2.selectedFeatures, accuracy)
-    );
+  calculatePotentialKValues(state) {
+    let datasetSize = state.data.length;
```
maureensturgeon, line 101: Nit: `trainingExamplesSize` and `possibleKValues` can both be consts because they aren't reassigned

maureensturgeon (approved):
I left a couple comments for further refactoring, but overall I think this is a great improvement!! :D thanks for sharing

breville (approved):
Thanks for the quick fix for small datasets!

PR #238, opened 2021-06-16
Strip leading and trailing spaces in categorical values

made-line:
In user-uploaded CSVs, categorical values with leading or trailing spaces were considered unique values along with its counterpart without the extra spaces, resulting in seemingly duplicated values. This change removes leading or trailing spaces from categorical values to prevent duplication. 

Example CSV:
Screen Shot 2021-06-16 at 3 42 36 PM

Before: 
Screen Shot 2021-06-16 at 3 59 51 PM

After: 
Screen Shot 2021-06-16 at 3 42 25 PM

### DIFF
--- a/src/csvReaderWrapper.js
+++ b/src/csvReaderWrapper.js
@@ -28,8 +28,17 @@ const isRowValid = (row) => {
   return cells.every(isCellValid)
 }
 
+const trimWhiteSpaces = (row) => {
+  var trimmedRows = {};
+  for (var cellName in row) {
+    trimmedRows[cellName] = row[cellName].trim();
+  }
+  return trimmedRows;
+}
+
 const cleanData = (data) => {
-  var cleanedData = data.filter(row => isRowValid(row));
+  var filterValidData = data.filter(row => isRowValid(row));
+  var cleanedData = filterValidData.map(row => trimWhiteSpaces(row));
   return cleanedData;
 }
 
--- a/src/csvReaderWrapper.js
+++ b/src/csvReaderWrapper.js
@@ -31,7 +31,11 @@ const isRowValid = (row) => {
 const trimWhiteSpaces = (row) => {
   var trimmedRows = {};
   for (var cellName in row) {
-    trimmedRows[cellName] = row[cellName].trim();
+    if (typeof row[cellName] === 'string') {
+      trimmedRows[cellName] = row[cellName].trim();
+    } else {
+      trimmedRows[cellName] = row[cellName]
+    }
   }
   return trimmedRows;
 }
--- a/src/csvReaderWrapper.js
+++ b/src/csvReaderWrapper.js
@@ -23,7 +23,7 @@ const cleanData = (data) => {
   var cleanedData = []
 
   for (var row of data) {
-    var cleanedRow = tryGetCleanedRow(row);
+    var cleanedRow = getCleanedRow(row);
     if (cleanedRow !== null) {
       cleanedData.push(cleanedRow);
     }
@@ -32,14 +32,16 @@ const cleanData = (data) => {
   return cleanedData;
 }
 
-const tryGetCleanedRow = (row) => {
-  for (var cellName in row) {
-    if (!isCellValid(row[cellName])) {
+const getCleanedRow = (row) => {
+  for (var column in row) {
+    var cellValue = row[column];
+
+    if (!isCellValid(cellValue)) {
       return null;
     }
 
-    if (typeof row[cellName] === 'string') {
-      row[cellName] = row[cellName].trim();
+    if (typeof cellValue === 'string') {
+      row[column] = cellValue.trim();
     }
   }
 
--- a/src/csvReaderWrapper.js
+++ b/src/csvReaderWrapper.js
@@ -40,9 +40,7 @@ const getCleanedRow = (row) => {
       return null;
     }
 
-    if (typeof cellValue === 'string') {
-      row[column] = cellValue.trim();
-    }
+    row[column] = cellValue.trim();
   }
 
   return row;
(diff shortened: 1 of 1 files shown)


src/csvReaderWrapper.js
```diff
@@ -28,8 +28,21 @@ const isRowValid = (row) => {
   return cells.every(isCellValid)
 }
 
+const trimWhiteSpaces = (row) => {
```
breville, line 31: supernit: whitespace is usually used as a collective noun so this can just be `trimWhitespace`.

src/csvReaderWrapper.js
```diff
@@ -28,8 +28,21 @@ const isRowValid = (row) => {
...
+  var trimmedRows = {};
```
breville, line 32: Since this function takes a single row, are we really generating multiple rows?

Erin007:
Is there a way to combine trimming white spaces with the iteration we're already doing? With this change we check _every single cell_ to see if it's empty and _every single cell_ to see if it needs to have whitespace trimmed. I wonder if there is a more performant approach that iterates through each cell only once?

src/csvReaderWrapper.js
```diff
@@ -19,23 +19,35 @@ export const parseCSV = (csvfile, download, useDefaultColumnDataType) => {
...
 
-const cleanData = (data) => {
-  var cleanedData = data.filter(row => isRowValid(row));
   return cleanedData;
 }
 
-const countRemovedRows = (originalData, cleanedData) => {
-  var removedRowsCount = originalData.length - cleanedData.length;
-  store.dispatch(setRemovedRowsCount(removedRowsCount));
+const tryGetCleanedRow = (row) => {
```
breville, line 35: nit: `try` kind of implies the use of exceptions (`try`/`catch`/etc.) so maybe just `getCleanedRow`?

src/csvReaderWrapper.js
```diff
@@ -19,23 +19,35 @@ export const parseCSV = (csvfile, download, useDefaultColumnDataType) => {
...
+  for (var cellName in row) {
+    if (!isCellValid(row[cellName])) {
+      return null;
+    }
+
+    if (typeof row[cellName] === 'string') {
```
breville, line 41: Pardon my ignorance, but are numerical cells actually numbers by this point?

src/csvReaderWrapper.js
```diff
@@ -19,23 +19,35 @@ export const parseCSV = (csvfile, download, useDefaultColumnDataType) => {
...
-const cleanData = (data) => {
-  var cleanedData = data.filter(row => isRowValid(row));
   return cleanedData;
 }
 
-const countRemovedRows = (originalData, cleanedData) => {
-  var removedRowsCount = originalData.length - cleanedData.length;
-  store.dispatch(setRemovedRowsCount(removedRowsCount));
+const tryGetCleanedRow = (row) => {
+  for (var cellName in row) {
```
Erin007, line 36: is `cellName` the column heading? In other places where we're accessing the value of a specific cell it's something like `cellValue = row[column]` I wonder if we want to do access the contents via row and column here too for consistency?

src/csvReaderWrapper.js
```diff
@@ -19,23 +19,35 @@ export const parseCSV = (csvfile, download, useDefaultColumnDataType) => {
...
+    if (!isCellValid(row[cellName])) {
```
Erin007, line 37: nit:`row[cellName]` is reused so can be extracted to a variable

src/csvReaderWrapper.js
made-line: Good point! Updated.

src/csvReaderWrapper.js
made-line: At this point, all values are converted to strings. I added this in because I accidentally uploaded an .xlsx file and got an error related to the type for `trim()`. I also thought it'd be good practice to program defensively since `trim()` can only be performed on a string. What do you think?

As a side note, I noticed .xls and .xlsx are supported file types for the file upload button in `SelectDataset`, but .xlsx files don't display the data correctly. Do we ever plan on supporting these files? The file upload button specifically mentions CSV, but just wondering if we should prevent accidental uploads of .xlsx files for users (like me).

src/csvReaderWrapper.js
made-line: That makes more sense, thanks!

src/csvReaderWrapper.js
made-line: Sounds good!

src/csvReaderWrapper.js
breville: It might be worth a comment explaining that this shouldn't ever be a string, otherwise it struck me that we might be expecting that to sometimes be the case here.

You make a really good point about non-CSV files, and something we should perhaps deal with a bit more proactively than we have.  It's out of scope for this PR, but should we try to catch non-CSV file issues earlier, rather than just deal with one of the symptoms in this cleanup function?  

And great catch about .xls & .xlsx files.  The former appear to be binary, and the latter a zip full of xml files.  We support neither, so we should remove them both from the upload accept filter.

breville (approved):
Thanks for addressing my concerns.  @Erin007 raised some great issues and should do a final review as well.

src/csvReaderWrapper.js
made-line: Looking at the Papa parse documentation, it looks like it always parses the values in to a string. @Erin007 can correct me if I'm wrong. I'll remove the string check since this seems redundant.

made-line: Also, thank you for looking in to the non-CSV file issue! I made a Trello card for it [here](https://trello.com/c/Pqn0zB6M/260-prevent-non-csv-file-uploads). I think it's a very small change, but I'll put it in its own PR to keep things organized.

Erin007 (approved):

dancodedotorg (approved):

PR #242, opened 2021-06-21
Tests for utils helper functions 

Erin007:


test/unit/utils.test.js
```diff
@@ -1,5 +1,71 @@
...
+    expect(result).toBe(false);
+  });
+});
+
+describe("getRandomInt", () => {
+  let max = 5;
+
+  test("random int is never greater than max", async () => {
+    const result = getRandomInt(max)
+    expect(result).toBeLessThan(max);
```
breville, line 63: So if things are wrong, this might fail, but only occasionally? :)

breville (approved):

test/unit/utils.test.js
Erin007: yes. I ran the test a handful of times and always got a valid output and that seemed better than no coverage, but yeah, testing randomness feel like a trap 😬

PR #243, opened 2021-06-21
Refactor to leverage selectors 

Erin007:
Our app has hundreds of lines of functions that look at the state tree and calculate, combine or manipulate new values based on the current state. Not all of the pieces of state are related to each other. For example, the imported data is not influenced in any way by whether instructions have been dismissed. However, we do many of these calculations and manipulations to determine derived state every time the state tree is updated. As a performance optimization, we can avoid unnecessary recalculations of derived state by leveraging [reselect](https://github.com/reduxjs/reselect), a library that allows us to customize the conditions under which selectors are called so that derived state is only recalculated when relevant pieces of state change. 

In this initial sample PR I extracted selectors for `getSelectedCategoricalFeatures` and `getSelectedNumericalFeatures` as a conversation starter to see if this felt like a good direction for further refactoring. 

Part way through these changes - `getSelectedCategoricalFeatures` was converted to a selector, but `getSelectedNumericalFeatures` was not yet - I went through the flow of AI Lab and selected both a numerical and categorical feature and logged each time these functions were called. 
Screen Shot 2021-06-21 at 3 19 54 PM

`getSelectedCategoricalFeatures` was efficiently called once, whereas `getSelectedNumericalFeatures` was called many times and continued to be unnecessarily called each time the state tree changed based on interactions with the Predict Panel.

Recalculating column type and which columns are selected isn't particularly expensive, but I think I like this pattern (so far) and am interested in pursuing this paradigm for more labor intensive calculations, such as those that iterate through all of the uploaded data (e.g. calculating the unique options in each categorical column).


package.json
```diff
@@ -75,6 +75,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.11.2",
     "@fortawesome/react-fontawesome": "^0.1.7",
     "chart.js": "^2.9.4",
-    "react-chartjs-2": "^2.11.1"
+    "react-chartjs-2": "^2.11.1",
+    "reselect": "^4.0.0"
```
breville, line 79: The `package-lock.json` changes look fairly significant; is it possible there is an earlier version of `reselect` that is more relevant to the other packages that we use?

breville:
This looks great!  This is definitely worth pursuing.

package.json
Erin007: Hmm. The latest release for reselect was in [version 4.0.0](https://github.com/reduxjs/reselect/releases) in Sept. 2018. It's possible there is an older version that wouldn't require bumping as many dependencies. I'm kind of tempted to bring everything else up to date. Stay tuned... going to crowdsource some opinions.

Erin007: aha! The package-lock.json file was so big because it was new! We didn't have a package-lock.json file before this change. I uninstalled and then re-installed reselect with yarn and the yarn.lock file looks completely reasonable now. 

Can you remind me if it was intentional to only manage packages with yarn rather than npm and if so, why?

Erin007:
Ok, I think this chunk of work is ready for review. I have some other ideas about how to continue to leverage selectors, but I think it would be good to roll out incrementally.

package.json
breville: Oh, my bad for not noticing that!  IIRC, we switched to yarn in the main repo long ago because we wanted a lock file, and at that time npm didn't offer that functionality.

src/redux.js
```diff
@@ -742,31 +720,7 @@ export function getSelectedColumnDescriptions(state) {
...
-export function getTableData(state, useResultsData) {
-  if (useResultsData) {
-    return getResultsDataInDataTableForm(state);
-  } else {
-    return state.data;
-  }
-}
-
-/* Function for retriving aggreate details about a currently selected column. */
+/* Functions for retriving aggregate details about a currently selected column. */
```
breville, line 723: While here, might as well fix the typo for "retrieving" as well :)

src/selectors.js
```diff
@@ -0,0 +1,118 @@
...
+    return extremaByColumn;
+  }
+)
+
+export const getExtremaCurrentColumn = createSelector(
+  [getCurrentColumn, getData],
+  (currentColumn, data) => {
+    return getExtrema(data, currentColumn);
+  }
+)
```
breville, line 118: I guess at some point we might have multiple files of selectors, grouped by category?  A single file seems fine for now, though.

test/unit/selectors.test.js
```diff
@@ -0,0 +1,118 @@
...
+  test("gets extrema current column", async () => {
+    const extrema = getExtremaCurrentColumn.resultFunc(
+      allNumericalState.currentColumn,
+      allNumericalState.data
+    )
+    expect(extrema.max).toBe(100);
+    expect(extrema.min).toBe(40);
+    expect(extrema.range).toBe(60);
+  });
+});
```
breville, line 118: Love all this testing!

breville (approved):
Pioneering stuff for us.  Overall, this looks like a great direction.  Nice work!

src/selectors.js
Erin007: Yeah, I imagine once I have the set of selectors written we might want to break them up kind of like I did for the redux file, but I'll do that in a future PR.

PR #244, opened 2021-06-21
Extract styles into their own file

Erin007:
No change for users, just moving style related code into a standalone `styles.js` file so it's easier to find when new folks come to the codebase. Glancing through the file, I think there's some vertical formatting tidying we can do and styles that we can delete since they no longer apply; I'll circle back for that clean up in a separate PR so the diff is easier to see.


src/styles.js
```diff
@@ -0,0 +1,912 @@
+export function getFadeOpacity(animationProgress) {
+  return animationProgress > 0.95
+    ? 1 - (animationProgress - 0.95) / 0.05
+    : animationProgress < 0.05
+    ? animationProgress / 0.05
+    : 1;
+}
+
```
breville, line 8: I wonder if this should go somewhere else?

src/UIComponents/SelectDataset.jsx
```diff
@@ -13,7 +13,7 @@ import {
 import { parseCSV } from "../csvReaderWrapper";
 import { parseJSON } from "../jsonReaderWrapper";
 import { allDatasets, getAvailableDatasets } from "../datasetManifest";
-import { styles } from "../constants";
+import { styles } from "../styles";
```
breville, line 16: Should we make `styles` the default export?

PR #248, opened 2021-06-23
Prevent non-CSV uploads

made-line:
Previously, we allowed uploads of `.xls` and `.xlsx` files which aren't supported. This PR specifies that we only accept .csv files.

For example, an `.xlsx` file like this:
Screen Shot 2021-06-23 at 1 15 32 PM

Renders like this in the data table:
Screen Shot 2021-06-23 at 1 12 58 PM

We currently use two ways to check for CSV files. One of the ways is built into `react-papaparse v3.8.0`, which is what we use to parse CSVs. The [documentation](https://react-papaparse.js.org/docs#local-files) mentions the default MIME type is for CSV: `'text/csv, .csv, application/vnd.ms-excel'`.  

Another way is through the file `<input>` element in `SelectDataset`. The `accept` attribute specifies the file type, which is changed to only `.csv` in this PR. According to [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept): 

> The accept attribute doesn't validate the types of the selected files; it provides hints for browsers to guide users towards selecting the correct file types. It is still possible (in most cases) for users to toggle an option in the file chooser that makes it possible to override this and select any file they wish, and then choose incorrect file types.
> 
> Because of this, you should make sure that expected requirement is validated server-side.

Since we don't send the CSV file itself to the server, we can add a validation in the `handleUploadSelect` function in `SelectDataset`. I'm wondering if this would be redundant since we already have two checks for the CSV file. 

If we wanted to add this validation, we'd also need to display a message somewhere (e.g., `DataCard` component). Let me know what you all think!

### DIFF
--- a/src/UIComponents/SelectDataset.jsx
+++ b/src/UIComponents/SelectDataset.jsx
@@ -13,7 +13,7 @@ import {
 import { parseCSV } from "../csvReaderWrapper";
 import { parseJSON } from "../jsonReaderWrapper";
 import { allDatasets, getAvailableDatasets } from "../datasetManifest";
-import { styles, fileTypeErrorMessage } from "../constants";
+import { styles } from "../constants";
 import ScrollableContent from "./ScrollableContent";
 
 class SelectDataset extends Component {
@@ -34,8 +34,7 @@ class SelectDataset extends Component {
     super(props);
 
     this.state = {
-      download: false,
-      showFileTypeErrorMessage: false
+      download: false
     };
   }
 
@@ -54,8 +53,7 @@ class SelectDataset extends Component {
       this.props.setSelectedCSV(csvPath);
       this.props.setSelectedJSON(jsonPath);
       this.setState({
-        download: true,
-        showFileTypeErrorMessage: false
+        download: true
       });
 
       parseCSV(csvPath, true, false);
@@ -66,19 +64,11 @@ class SelectDataset extends Component {
 
   handleUploadSelect = event => {
     this.props.resetState();
-    const selectedCSV = event.target.files[0]
-    this.props.setSelectedCSV(selectedCSV);
+    this.props.setSelectedCSV(event.target.files[0]);
     this.setState({
       download: false
     });
-    if (selectedCSV.name.includes(".csv") ||
-    selectedCSV.type === "text/csv") {
-        parseCSV(event.target.files[0], false, true);
-    } else {
-      this.setState({
-        showFileTypeErrorMessage: true
-      });
-    }
+    parseCSV(event.target.files[0], false, true);
   };
 
   render() {
@@ -132,7 +122,7 @@ class SelectDataset extends Component {
               <input
                 className="csv-input"
                 type="file"
-                accept=".csv"
+                accept=".csv,text/csv,application/vnd.ms-excel"
                 name="file"
                 placeholder={null}
                 onChange={this.handleUploadSelect}
@@ -140,11 +130,6 @@ class SelectDataset extends Component {
                 value=""
               />
             </label>
-            {this.state.showFileTypeErrorMessage && (
-              <span style={styles.fileTypeErrorMessageConainer}>
-                {fileTypeErrorMessage}
-              </span>
-            )}
           </div>
         )}
       </div>
--- a/src/UIComponents/SelectDataset.jsx
+++ b/src/UIComponents/SelectDataset.jsx
@@ -122,7 +122,7 @@ class SelectDataset extends Component {
               <input
                 className="csv-input"
                 type="file"
-                accept=".csv,.xls,.xlsx"
+                accept=".csv"
                 name="file"
                 placeholder={null}
                 onChange={this.handleUploadSelect}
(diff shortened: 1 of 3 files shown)


dancodedotorg:
👋 I missed the chance to comment on this in the other PR, but wanted to add: _**thank you**_ for adding this functionality. Just working with CSVs is the goal because (1) we can't assume every school/student has Excel, (2) AppLab's data tab exports to CSV, and (3) CSV is the most common filetype for data science and ML, so let's get students used to it. 🎉🎉🎉🎉

made-line:
Thank you for the clarification, @dancodedotorg! Those are great points :smiley_cat:

breville:
Since the MIME type of `application/vnd.ms-excel` is [for](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) an `.xls` file, it would be worth disabling.

That said, these mitigations don't seem to prevent uploading & parsing of a non-CSV file, so it would be worth investigating whether it might be possible to do more.  The idea is discussed [here](https://github.com/mholt/PapaParse/issues/501), though without much of a solution, except the idea that they can look for a specific header to reject a known non-CSV file type.  But maybe we could create a heuristic that at least improves the odds of it being a valid CSV file, such as only containing some kind of human-readable text as opposed to binary data?

Also, there do appear to be some open source CSV validator modules that might be useful, such as [this one](https://github.com/zazuko/csv-validate), but not sure whether they give us more over what we already have, nor whether it's worth more investment at this point.

src/UIComponents/SelectDataset.jsx
```diff
@@ -64,11 +66,19 @@ class SelectDataset extends Component {
...
     this.props.resetState();
-    this.props.setSelectedCSV(event.target.files[0]);
+    const selectedCSV = event.target.files[0]
+    this.props.setSelectedCSV(selectedCSV);
     this.setState({
       download: false
     });
-    parseCSV(event.target.files[0], false, true);
+    if (selectedCSV.name.includes(".csv") ||
+    selectedCSV.type === "text/csv") {
```
breville, line 75: Unfortunately, it sounds like this change might actually cause unnecessary failures on some systems: https://github.com/mholt/PapaParse/issues/18

breville:
I think it's possible to modify the `accept` property as described at https://react-papaparse.js.org/docs#local-files

However, as noted in a comment, it sounds like some systems get an unexpected MIME type for CSV files due to an application overriding it, so this might not be a good idea after all.  (https://github.com/mholt/PapaParse/issues/18)

In which case, as you say, we can wait until we get more feedback.

made-line:
> I think it's possible to modify the `accept` property as described at https://react-papaparse.js.org/docs#local-files

The documentation defines it, but I don't see a way to configure it in the examples or in the [DEFAULT CONFIG WITH ALL OPTIONS](https://react-papaparse.js.org/docs#config-default) section. Can you point it out in the docs?
 
> However, as noted in a comment, it sounds like some systems get an unexpected MIME type for CSV files due to an application overriding it, so this might not be a good idea after all. ([mholt/PapaParse#18](https://github.com/mholt/PapaParse/issues/18))

If this is the case then we shouldn't try to change the `accept` property to exclude `application/vnd.ms-excel`. The changes in this PR also don't exclude `application/vnd.ms-excel`. Files could have a `application/vnd.ms-excel` type and a `.csv` extension and they'd still proceed through the workflow. It checks for a `.csv` __or__ a CSV MIME type.

breville:
Oh, the `accept` type is on the `input` receiving the upload.  In their documentation, it's wrapped by `CSVReader` and passed through, but we apparently went straight to an `input`.  

So, after all this, I think we should just update our `input` to match PapaParse's default: 
```
'text/csv, .csv, application/vnd.ms-excel'
```
from https://github.com/Bunlong/react-papaparse/blob/89fae2832f26cceed74d6d2427b01fdadcd83393/src/CSVReader.tsx#L14

and not test anything additional in code.  We want to include the third MIME type for the reasons detailed in the linked discussion: on some systems this MIME type represents a valid CSV.

made-line:
Sounds good. Thank you for looking in to this!

made-line:
Circling back to the discussion about string validation in https://github.com/code-dot-org/ml-playground/pull/238: when users select an `.xls` file, the `DataCard` component doesn't render and there's a console error: 
Screen Shot 2021-07-02 at 1 53 02 PM

The user is then stuck there with no indication of an issue with the file. Adding in the string check in `csvReaderWrapper.js` removes the error and the user's able to continue to the next page to see the data is rendered incorrectly. Should we address this and how so? Should I add the string check back in or should we catch an exception for that error?

breville:
Per offline discussion, I think it's worth guarding against the non-string case in `isCellValid` as you've done.

That said, I dug into why we were getting a non-string, and it's because of this feature: https://www.papaparse.com/docs#data

It would be worth guarding against that.  Perhaps with an extra condition like this?

```
const getCleanedRow = (row) => {
  for (var column in row) {
    if (column !== "__parsed_extra") {
      var cellValue = row[column];

      if (!isCellValid(cellValue)) {
        return null;
      }

      row[column] = cellValue.trim();
    }
  }

  return row;
}
```

made-line:
Good find! Thank you for your guidance.

breville (approved):

PR #251, opened 2021-06-24
ColumnInspector refactor

Erin007:
We had a fairly large, complex component `ColumnInspector` that was handling many aspects of displaying the details of a selected column. I extracted 5 smaller child components that each handle a specific part of the functionality: `ColumnDetailsCategorical`, `ColumnDetailsNumerical`, `AddFeatureButton`, `SelectLabelButton` and `ColumnDataTypeDropdown`. Now each component only receives the state they need and the code is more modular. This change has the additional benefit of hoisting logic out of the UI components and updating `getCurrentColumnData` to be more efficient; for example, we were previously calling `getUniqueOptions` for numerical columns which is excessive and unnecessary.

### DIFF
--- a/src/UIComponents/AddFeatureButton.jsx
+++ b/src/UIComponents/AddFeatureButton.jsx
@@ -12,9 +12,9 @@ class AddFeatureButton extends Component {
     addSelectedFeature: PropTypes.func.isRequired
   };
 
-  addFeature = e => {
-    this.props.addSelectedFeature(this.props.column);
-    e.preventDefault();
+  addFeature = (event, column) => {
+    this.props.addSelectedFeature(column);
+    event.preventDefault();
   };
 
   render() {
@@ -23,7 +23,7 @@ class AddFeatureButton extends Component {
     return (
       <button
         type="button"
-        onClick={e => this.addFeature(e, column)}
+        onClick={event => this.addFeature(event, column)}
         style={styles.selectFeaturesButton}
       >
         Add feature
--- a/src/UIComponents/AddFeatureButton.jsx
+++ b/src/UIComponents/AddFeatureButton.jsx
@@ -1,0 +1,43 @@
+/* React component to handle selecting columns as features. */
+import PropTypes from "prop-types";
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import { styles } from "../constants";
+import { addSelectedFeature } from "../redux";
+
+
+class AddFeatureButton extends Component {
+  static propTypes = {
+    column: PropTypes.string,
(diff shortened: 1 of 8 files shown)


Erin007:
@maureensturgeon @cforkish @davidsbailey this is the component I mentioned at last book club 😄

src/UIComponents/AddFeatureButton.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { addSelectedFeature } from "../redux";
+
+
+class AddFeatureButton extends Component {
+  static propTypes = {
+    column: PropTypes.string,
+    addSelectedFeature: PropTypes.func.isRequired
+  };
+
+  addFeature = e => {
```
breville, line 15: nit: the next file uses `event`.  Should we consistently use `e` or `event` across the codebase?

src/UIComponents/ColumnDetailsCategorical.jsx
```diff
@@ -0,0 +1,78 @@
...
+        <div style={styles.bold}>Column information:</div>
+
+        <div style={styles.barChart}>
+          {barData.labels.length <= maxLabelsInHistogram && (
+            <Bar
+              data={barData}
+              width={100}
+              height={150}
+              options={chartOptions}
+            />
```
breville, line 65: Maybe out of scope here, and essentially a preexisting issue that I always meant to circle back to, but it's a bit inconsistent that `ColumnDetailsCategorical` contains the bar chart component, but `ColumnDetailsNumerical` doesn't contain the scatter plot component.

Erin007: I don't think `BarChart` is to `ColumnDetailsCategorical` as `ScatterPlot` is to `ColumnDetailsNumerical`. We show a bar chart as a visualization of the values in a single categorical column. The equivalent information for a single numerical column is max/min/range. Scatterplot shows the relationship between 2 numerical columns similar to how the cross tab shows the relationship between two categorical columns. `ScatterPlot` and `CrossTab` are already fairly tidy in that they are their own components that each get the information they need directly from redux so I do plan to touch then as part of this change.

src/UIComponents/AddFeatureButton.jsx
Erin007: sure. I'll go with "event" since it's easier to search unless you have a preference.

breville: Sounds good.

breville (approved):

maureensturgeon (commented):
Such a great change, thanks for sharing!

PR #253, opened 2021-06-25
[Datasets] Shoes + punctuation

dancodedotorg:
I thought I was done uploading datasets, but then I realized I needed one more to record a video with and I couldn't use one I already planned for the curriculum. So, this PR includes:

- A new dataset about shoes
- Brendan noticed (rightfully so) that I keep forgetting to include periods in my json metadata, which I realized is probably something I've been doing all along. So - I went back and added periods to all of my other datasets and some of our existing datasets too. Now we're super consistent about punctuation =)


public/datasets/billionaires.json
```diff
@@ -85,17 +85,17 @@
...
       "type": "categorical",
       "id": "Wealth Industry",
-      "description": "The industry their wealth is in"
+      "description": "The industry their wealth is in."
     },
     {
       "type": "categorical",
       "id": "Wealth Inherited",
-      "description": "Whether their wealth was inherited and where from, e.g. 'not inherited', 'father', '3rd generation'"
+      "description": "Whether their wealth was inherited and where from, e.g. 'not inherited', 'father', '3rd generation'."
```
breville, line 98: We are a little inconsistent about using single- or double-quotes.

public/datasets/shoe_survey_toy.json
```diff
@@ -0,0 +1,46 @@
...
+    },
+    {
+      "id": "notice shoes",
+      "type": "categorical",
+      "description": "Do you care if people notice your shoes?"
+    },
+    {
+      "id": "shoe",
+      "type": "categorical",
+      "description": "What kind of shoe are you planning on wearing today?"
```
breville, line 42: shoes?

breville (approved):

PR #256, opened 2021-06-29
Save model form and Model Card updates 

Erin007:
Re-ordered sections of the Save Model form and the Model Card and updated a few strings per [request](https://docs.google.com/document/d/1GLDEz5Nb6SYqTohXU992fAxS-D0Lg8cQA5ic9dJCNXA/edit) from the Curriculum team. I also widened the column description entry boxes to match other save form fields. 

Save form
Screen Shot 2021-06-29 at 1 34 26 PM
Screen Shot 2021-06-29 at 3 25 15 PM

Summary screen 
Screen Shot 2021-06-29 at 3 13 48 PM
Screen Shot 2021-06-29 at 3 13 54 PM

### DIFF
--- a/src/UIComponents/ModelCard.jsx
+++ b/src/UIComponents/ModelCard.jsx
@@ -45,21 +45,39 @@ class ModelCard extends Component {
         <div style={styles.modelCardContainer}>
           <h3 style={styles.modelCardHeader}>{trainedModelDetails.name}</h3>
           <div style={styles.modelCardSubpanel}>
-            <h5 style={styles.modelCardHeading}>Summary</h5>
+            <h5 style={styles.modelCardHeading}>Accuracy</h5>
             <div style={styles.modelCardContent}>
               <p style={styles.modelCardDetails}>
-                Predict {label.id} based on{" "}
-                {selectedFeatures.length > 0 && percentCorrect && (
+                {percentCorrect && (
                   <span>
-                    {selectedFeatures.join(", ")} with {percentCorrect}%
-                    accuracy.
+                    {percentCorrect}%
                   </span>
                 )}
               </p>
             </div>
           </div>
           <div style={styles.modelCardSubpanel}>
-            <h5 style={styles.modelCardHeading}>About the Data </h5>
+            <h5 style={styles.modelCardHeading}>Intended Use</h5>
+            <div style={styles.modelCardContent}>
+              {trainedModelDetails.potentialUses && (
+                <p style={styles.modelCardDetails}>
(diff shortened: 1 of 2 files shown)


dancodedotorg (approved):
🎉🎉🎉 Looks great! Thank you!

src/UIComponents/ModelCard.jsx
```diff
@@ -45,52 +45,63 @@ class ModelCard extends Component {
         <div style={styles.modelCardContainer}>
           <h3 style={styles.modelCardHeader}>{trainedModelDetails.name}</h3>
           <div style={styles.modelCardSubpanel}>
-            <h5 style={styles.modelCardHeading}>Summary</h5>
+            <h5 style={styles.modelCardHeading}>Accuracy</h5>
             <div style={styles.modelCardContent}>
               <p style={styles.modelCardDetails}>
-                Predict {label.id} based on{" "}
-                {selectedFeatures.length > 0 && percentCorrect && (
+                {percentCorrect && (
```
breville, line 51: What happens if it's `0%`?  Is there a chance we want to show it but it won't show?

src/UIComponents/ModelCard.jsx
```diff
@@ -45,52 +45,63 @@ class ModelCard extends Component {
...
+            <h5 style={styles.modelCardHeading}>About the Data</h5>
             <div style={styles.modelCardContent}>
-              {trainedModelDetails.potentialMisuses && (
+              {datasetDetails.description && (
                 <p style={styles.modelCardDetails}>
-                  {trainedModelDetails.potentialMisuses}
+                  {datasetDetails.description}
                 </p>
               )}
+              {datasetDetails.numRows && (
```
breville, line 87: Similarly here, what if there are `0` rows?

Erin007: There's no way to successfully train a model and get to the summary screen if there are 0 rows of data so I'm not too worried about this case.

src/UIComponents/ModelCard.jsx
Erin007: Screen Shot 2021-06-29 at 9 09 47 PM

This is kind of interesting... in the case where `percentCorrect` === 0 the zero will render, but not 0%, which I think is probably fine. It's rare to have a 0% accurate model, and even more rare that someone would want to save a 0% accurate model because it'd be useless. _But_ I also trust that this component will always have `percentCorrect` when needed so it seems ok to remove the check for `percentCorrect` and just render the value.

breville (approved):

PR #254, opened 2021-06-28
Current column selectors 

Erin007:
Continuation of #251 and #243 

- extracts `UniqueOptionsWarning` from `ColumnInspector` 
- connects `ColumnDetailsCategorical` and `ColumnDetailsNumerical`
- changes "currentColumnData" to "currentColumnDetails"  in `ColumnInspector` to avoid confusion with the actual data from the dataset stored in a column 
- moves a bunch of redux and columnDetails functions into the new `currentColumnsSelectors` file 
- initial tests for `currentColumnsSelectors`

PR #261, opened 2021-06-30
Fix manual column descriptions for user-uploaded CSV

breville:
This should fix the issue described in https://github.com/code-dot-org/ml-playground/pull/259 which was a regression introduced in https://github.com/code-dot-org/ml-playground/pull/254.

There were three things causing text entered in the "Column Descriptions" fields to not be visible to the user:

- `getColumnDescription` had a collision between two `column` variables.
- The `SET_TRAINED_MODEL_DETAIL` action needed to make a fresh copy of the `trainedModelDetails` so that the memoized selector would notice that a value inside of it had changed, per [this advice](https://github.com/reduxjs/reselect#q-why-isnt-my-selector-recomputing-when-the-input-state-changes).
- The `SET_TRAINED_MODEL_DETAIL` action needed to set the `trainedModelDetails` field when returning the new state.  (Interestingly, this issue had been around since the code was first written five months ago, but not noticed because it was also modifying the `trainedModelDetails` value in-place!)

### DIFF
--- a/src/helpers/columnDetails.js
+++ b/src/helpers/columnDetails.js
@@ -61,11 +61,11 @@ export function containsOnlyNumbers(data, column) {
   return columnData.every(cell => !isNaN(cell));
 }
 
-export function getColumnDescription(column, metadata, trainedModelDetails) {
+export function getColumnDescription(columnId, metadata, trainedModelDetails) {
   // Use metadata if available.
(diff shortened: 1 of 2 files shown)


Erin007 (approved):
ooooh!! Thanks for digging in and putting these pieces together.

PR #271, opened 2021-07-01
Predict input box now accepts more than one digit

Erin007:


src/UIComponents/Predict.jsx
```diff
@@ -30,7 +30,7 @@ class Predict extends Component {
   };
 
   handleChange = (event, feature) => {
-    const testData = this.props.testData;
+    const testData = {...this.props.testData};
```
breville, line 33: Ah, without this, React doesn't think a prop has changed, and doesn't re-render.

breville (approved):
Thanks for prepping this.

For consideration, https://github.com/code-dot-org/ml-playground/pull/272 should do the same thing, but moves more of the business logic into the redux action, so that the UI code doesn't have to know quite as much about what's going on.

Erin007:
Closing in favor of #272

PR #272, opened 2021-07-01
Predict input box now accepts more than one digit

breville:
An alernate possibility to https://github.com/code-dot-org/ml-playground/pull/271, which puts more of the business logic into the redux action.


Erin007 (approved):
Thanks for adjusting this! I'm in favor of more logic in the reducer rather than the component, so I'll close #272

PR #274, opened 2021-07-12
Supplement UI test for training a model

made-line:
https://github.com/code-dot-org/code-dot-org/pull/41527


src/UIComponents/DataTable.jsx
```diff
@@ -134,6 +134,7 @@ class DataTable extends Component {
                   return (
                     <td
                       key={columnId}
+                      className="table-column"
```
breville, line 137: Maybe `"data-table-column"` since we also use HTML tables in other places too?

made-line: Sounds good!

Erin007 (commented):
There's a convention in the main repo ([for example](https://github.com/code-dot-org/code-dot-org/search?q=uitest)) to prepend `uitest` to selectors that are used specifically for UI tests. Save future folks from going hunting for css and it's easier to search. It'd be good to be consistent here. 

Otherwise, excited to see this work coming along! Yay for test coverage!

made-line:
@Erin007 Thanks for pointing that out!

Erin007 (approved):

PR #276, opened 2021-07-13
Accessibility improvements 

Erin007:
I incorporated [jsx-a11y](https://www.npmjs.com/package/eslint-plugin-jsx-a11y) to include accessibility rules in our linting. Using those rules, I added alt text to images and added roles and `onKeyDown` handlers for visible, interactive elements.

### DIFF
--- a/src/UIComponents/GenerateResults.jsx
+++ b/src/UIComponents/GenerateResults.jsx
@@ -9,6 +9,7 @@ import aiBotBody from "@public/images/ai-bot/ai-bot-body.png";
 import blueScanner from "@public/images/ai-bot/blue-scanner.png";
 import background from "@public/images/results-background-light.jpg";
 import DataTable from "./DataTable";
+import { TestingAnimationDescription } from "./AnimationDescriptions";
 
 const framesPerCycle = 80;
 const maxNumItems = 7;
@@ -202,6 +203,7 @@ class GenerateResults extends Component {
             </div>
           </div>
         </div>
+        <TestingAnimationDescription />
       </div>
     );
   }
--- a/src/UIComponents/GenerateResults.jsx
+++ b/src/UIComponents/GenerateResults.jsx
@@ -184,12 +184,12 @@ class GenerateResults extends Component {
             <img
               src={aiBotHead}
               style={styles.trainBotHead}
-              alt="A.I. Bot head"
+              alt="A.I. bot head"
             />
             <img
               src={aiBotBody}
               style={styles.trainBotBody}
-              alt="A.I. Bot body"
+              alt="A.I. bot body"
             />
             <div
               style={{ width: 150, position: "absolute", top: 140, zIndex: -1 }}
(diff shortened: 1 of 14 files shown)


src/UIComponents/GenerateResults.jsx
```diff
@@ -181,14 +181,23 @@ class GenerateResults extends Component {
...
+            <img
+              src={aiBotHead}
+              style={styles.trainBotHead}
+              alt="A.I. Bot head"
+            />
+            <img
+              src={aiBotBody}
+              style={styles.trainBotBody}
+              alt="A.I. Bot body"
+            />
```
breville, line 193: I'm a bit torn about including "Bot" because of our [previous naming discussions](https://codedotorg.slack.com/archives/C016VGH1BT6/p1616203569028600?thread_ts=1615308631.003000&cid=C016VGH1BT6).  Maybe "bot" to be descriptive without formally naming it?

src/UIComponents/SelectDataset.jsx
```diff
@@ -115,6 +115,7 @@ class SelectDataset extends Component {
                     style={styles.selectDatasetImage}
                     draggable={false}
                     className="ailab-image-hover"
+                    alt={dataset.name}
```
breville, line 118: I'm not sure the right practice, but is it possible we should add some context as well, like "Select [dataset.name]" or "Dataset related to [dataset.name]", etc.?

Erin007: I read a few articles and docs. There doesn't seem to be a clear definitive answer. In some cases it's considered redundant to repeat instructions for similar elements as long as the context can be inferred from other surrounding elements. In other cases, the recommendation is that more description and detail in an alt tag is preferred. 🤷‍♀️

epeach:
Me, being psyched about this: 
excited dog

src/UIComponents/GenerateResults.jsx
```diff
@@ -181,14 +181,23 @@ class GenerateResults extends Component {
           }}
         >
           <div className="ailab-image-hover" style={styles.trainBot}>
-            <img src={aiBotHead} style={styles.trainBotHead} />
-            <img src={aiBotBody} style={styles.trainBotBody} />
+            <img
+              src={aiBotHead}
+              style={styles.trainBotHead}
+              alt="A.I. bot head"
```
epeach, line 187: Alt tags are definitely an art, not a science, so this is just a suggestion, but I wonder if there is a way for us to just summarize the three "A.I bot head/body/scanner beam" alt-texts to have just one alt-text that's like "A.I Bot with Scanner Beam" or something like that. From my point of view, that would summarize the idea of what is happening similar to how a sighted person would take in all three pieces as a whole.

Erin007:
I just added in animation descriptions based on the [advice here](https://www.savaslabs.com/blog/alt-text-accessible-animations). What do you think? @epeach Does this help demonstrate cohesively what's happening with all the bot parts concern you have?

epeach:
That's awesome! I'd never seen that system before, but it looks like exactly what we want here.

breville (approved):
Great work!

PR #277, opened 2021-07-16
Document PropType shapes for results

Erin007:
This is the initial work to document the expected PropTypes for components with nested or complex state to make the codebase more readable and easier to modify over the long term. Specifically this PR defines a `resultsPropType` shape used by the `ResultsTable` and `ResultsDetails` component.


Erin007:
@BennettElisa  https://reactjs.org/docs/typechecking-with-proptypes.html and the first half of this [article](https://medium.com/@duhseekoh/shaping-up-your-react-components-with-proptypes-shape-and-selectors-9d1111d0566f) provide more context

breville (approved):

PR #285, opened 2021-08-14
React PropTypes for categoricalColumnDetailsShape

BennettElisa:


BennettElisa:
Moved to different branch

PR #291, opened 2022-10-08
Adding support for i18n

daynew:
Laying the foundation for internationalization(i18n) support in ml-playground. This adds `i18n.js` which provides the `I18n.t` API for getting translated content. For example, calling `I18n.t('trainModelButtonText')` returns "Train" which is a string stored in `i18n/ailab.json`

This implementation follows the same pattern as [AI for Oceans](https://github.com/code-dot-org/ml-activities)

`i18n/ailab.json` will store key/value pairs where the key is a unique identifier for a string in the AI Lab UI, and the value is a MessageFormatted string to display in the UI. We use MessageFormat so we can support strings with dynamic content. See the unit tests for an example string.

## Links
* [JIRA](https://codedotorg.atlassian.net/browse/FND-2082)

## Testing done
* Manual testing done to change the "Train" button to be "!Train". See screenshot below
* Added unit tests.

## Screenshot
Note the button text "!Train". This translation was manually added to verify that the `I18n` library was being used by the frontend.
Screen Shot 2022-10-07 at 5 30 07 PM

## Follow-up work
* Separate PR for making the rest of the UI strings translatable.
* Separate PR for adding `ailab.json` to the code-dot-org i18n sync and getting it translated in Crowdin.

### DIFF
--- a/src/helpers/navigationValidation.js
+++ b/src/helpers/navigationValidation.js
@@ -1,3 +1,4 @@
+import I18n from '../i18n'
 /*
 Validation checks to determine if app set up is ready for machine learning
 training. Panels prompt users to incrementally complete actions in preparation
@@ -175,7 +176,7 @@ export function prevNextButtons(state) {
       ? { panel: "dataDisplayLabel", text: "Back" }
       : null;
     next = isPanelAvailable(state, "trainModel")
-      ? { panel: "trainModel", text: "Train" }
+      ? { panel: "trainModel", text: I18n.t('trainModelButtonText') }
       : null;
   } else if (state.currentPanel === "trainModel") {
     if (state.trainedModel) {
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ import { allDatasets } from "./datasetManifest";
 import { parseCSV } from "./csvReaderWrapper";
 import { parseJSON } from "./jsonReaderWrapper";
 import { TestDataLocations } from "./constants";
+import I18n from './i18n';
 
 export const store = createStore(rootReducer);
 
@@ -25,6 +26,7 @@ let saveTrainedModel = null;
 let onContinue = null;
 
 export const initAll = function (options) {
+  I18n.initI18n(options.i18n);
   // Handle an optional mode.
   const mode = options && options.mode;
   onContinue = options && options.onContinue;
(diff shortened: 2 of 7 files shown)


src/helpers/navigationValidation.js
```diff
@@ -175,7 +176,7 @@ export function prevNextButtons(state) {
       ? { panel: "dataDisplayLabel", text: "Back" }
       : null;
     next = isPanelAvailable(state, "trainModel")
-      ? { panel: "trainModel", text: "Train" }
+      ? { panel: "trainModel", text: I18n.t('trainModelButtonText') }
```
annaxuphoto, line 179: is this the only string we're trying to translate? or is this PR just to implement functionality?

src/index.js
```diff
@@ -18,13 +18,15 @@ import { allDatasets } from "./datasetManifest";
...
 import { TestDataLocations } from "./constants";
+import I18n from './i18n';
 
 export const store = createStore(rootReducer);
 
 let saveTrainedModel = null;
 let onContinue = null;
 
 export const initAll = function (options) {
+  I18n.initI18n(options.i18n);
```
annaxuphoto, line 29: curious what `options.i18n` is/where it comes from

annaxuphoto (approved):
looks good to me!

src/index.js
wilkie: And does the `initAll` call in `indexDev.js` need a reference to `i18n`?

src/helpers/navigationValidation.js
daynew: The goal of this PR is to just establish an i18n foundation. A followup PR will have the rest of the strings.

src/index.js
daynew: > curious what `options.i18n` is/where it comes from

This will be passed in `Ailab.js` in the code-dot-org project. All of our translations will live in the code-dot-org project and be passed along using `options.i18n`. You can see that happening here: https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/ailab/Ailab.js#L206

> And does the initAll call in indexDev.js need a reference to i18n?
It doesn't need it because `indexDev.js` calls `index.js:initAll`. `index.js:initAll` doesn't need `i18n` provided because it will always default to English. It only needs i18n to be defined if we want to override the English with a different locale.
I will add this documentation.

wilkie: That makes sense! Thanks.

PR #292, opened 2022-10-18
Add i18n for most UI strings in AI Lab

daynew:
This PR makes most of the UI text in AI Lab be translatable.
* [JIRA](https://codedotorg.atlassian.net/browse/FND-2082)

### Formatting changes
Changes I've made that aren't strictly related to translating text, but I think are more universal / simple and will be easier for all students to understand

Changing `A: B`
It many places in the code I opted to remove or replace colons (`:`) placed between a label and text because the character is being used for structural purposes and may not be common in other languages. For example
```
Last updated: 1997
```
becomes
```
Last updated
1997
``` 

Removing commas from lists of items.
Not every languages uses commas for list of items, so I have changed the UI to instead separate items by whitespace instead.
```
apples, oranges, lemons, peaches
```
becomes
```
apples oranges lemons peaches
```

## How to review this PR
There are a lot of changes in this PR, so here are some steps you can take to get started:
1. Read through all the strings in `ailab.json`. These are all the English strings which can be translated. Does something look wrong? Is there a string missing?
1. Check out the "Formatting changes" section above to see if they are acceptable.
1. Watch the demo video below to see where in the UI the strings are used.
1. Look through each individual commit so you can see a smaller set of changes.
1. Review the "Future work" below and note if there is anything else which should be done.

## Demo
It would be way too many screenshots so I have opted for a video. Every string with a `!` at the beginning is a string which can be translated. The exclamation point is a manual change I am doing to demo this change.

https://user-images.githubusercontent.com/1372238/196313540-435f0816-7d3a-41b0-8469-331681f94b81.mov

## Future work
* A different PR will address the  "Predict ___ based on ____" strings.
* A different PR will address translating the CSV content.


src/UIComponents/Predict.jsx
```diff
@@ -112,12 +115,13 @@ class Predict extends Component {
                 className="ailab-image-hover"
                 style={styles.predictBot}
                 src={aiBotBorder}
-                alt="A.I. bot"
+                alt={I18n.t("aiBotAltText")}
```
Erin007, line 118: Thanks for thinking about accessibility here 🎉

src/UIComponents/DataCard.jsx
```diff
@@ -23,57 +24,58 @@ class DataCard extends Component {
...
+        I18n.t("dataCardWarningLargeDataset", {"rowCount": dataLengthLimit})
       );
     }
 
     const removedRowsMsg =
       removedRowsCount > 0
-        ? removedRowsCount === 1
-          ? `${this.props.removedRowsCount} row was removed because it contained an empty cell.`
-          : `${this.props.removedRowsCount} rows were removed because they contained empty cells.`
+        ? I18n.t("dataCardRemovedRows", {"rowCount": removedRowsCount})
```
breville, line 33: We lose the "1 row" case?

wilkie: I am curious how crowdin would handle the pluralization cases... because you would normally do something like (Rails i18n in my example ... [messageformat has a different take](https://messageformat.github.io/messageformat/guide/#pluralformat)):

```
dataCardRemovedRows:
  one: "{count} row removed because it contained empty cells.",
  other: "{count} rows removed because they contained empty cells.",
```

And then `I18n.t("dataCardRemovedRows", {"count": removedRowsCount})`

src/UIComponents/Predict.jsx
```diff
@@ -51,12 +52,14 @@ class Predict extends Component {
...
-                    : &nbsp;
+                    &nbsp;
                     <input
                       type="number"
                       onChange={event => this.handleChange(event, feature)}
                       value={this.props.testData[feature] || ""}
-                      placeholder={`min: ${+min}, max: ${+max}`}
+                      placeholder={I18n.t(
+                        "predictPlaceholder",
+                        {"minimum": min, "maximum": max})}
```
breville, line 62: I think we might want to keep the `+`s, per [this comment](https://github.com/code-dot-org/ml-playground/pull/122#discussion_r603569945).

src/csvReaderWrapper.js
```diff
@@ -56,17 +59,20 @@ const isCellValid = (cell) => {
...
   var cleanedData = cleanData(data);
+
+  if (cleanedData.length < MIN_CSV_ROWS) {
+    store.dispatch(setInvalidData("tooFewRows"));
+    return;
+  } else if (Object.keys(cleanedData[0]).length < MIN_CSV_COLUMNS) {
+    store.dispatch(setInvalidData("tooFewColumns"));
+    return;
+  }
+
```
breville, line 70: Just curious, is this a bug fix?

breville (approved):
This is fantastic.  Genuinely appreciate the amount of work here, and the thoughtfulness throughout.

Left a couple inline comments.  Also curious whether the app will need other changes to support RTL layouts?

Was also wondering whether we should lightly bold the new headings which were previous `:` prompts, but given the current hierarchy of that content, probably not necessary.

src/UIComponents/Predict.jsx
wilkie: That's true... `min` is a fixed number so it would be printed with trailing zeroes to denote its significance. The `+` would essentially perform a `parseFloat` and convert it back to a float where it would be printed without them.

src/App.js
```diff
@@ -10,7 +10,7 @@ import Results from "./UIComponents/Results";
 import Predict from "./UIComponents/Predict";
 import SaveModel from "./UIComponents/SaveModel";
 import ModelCard from "./UIComponents/ModelCard";
-import { styles, saveMessages } from "./constants";
+import { styles} from "./constants";
```
wilkie, line 13: Perhaps just an absolutely pedantic code style update to add the space before the brace.

wilkie (approved):
I did not run this locally, but it looks very good! I appreciate the care and attention to removing the colons and English-centric punctuation.

smusoke (approved):
Also looked this over and looks great. The only questions I had were already captured by other reviewers.

src/UIComponents/DataCard.jsx
daynew: You are right @wilkie, I tried to do that but this repo is using an old version of `messageformat` which doesn't have that API. When I tried to upgrade `messageformat` I got some errors I didn't know how to deal with so I did a quick workaround. I will work on updating `messageformat` and doing what you suggested.

src/csvReaderWrapper.js
daynew: Yes, if there were 0 rows or columns, then exceptions would be thrown because `countRemovedRows` has some code which tried to reference the first row.

I think this structure makes more sense anyways because you usually want to check you have valid inputs before running any calculations on them.

daynew:
> I'm popping in because I'm curious about international dataset curation - will Code.org only offer the English, US-centric sets we have now? Folks can always upload their own data in whatever language they chose, but it was a bit of an undertaking to find datasets that had potentially interesting findings and I don't know what's out there in other languages.

@Erin007 , non-US-centric datasets are out of scope for this push to make CSD translatable, but we agree we need to diversify our datasets. Many of these datasets are not interesting to students outside the US.  That will be something the Student Experience team will need to resolve.

src/App.js
daynew: fixed

src/UIComponents/Predict.jsx
daynew: fixed

src/UIComponents/DataCard.jsx
daynew: The package has been updated to support plurals.

daynew:
> Left a couple inline comments. Also curious whether the app will need other changes to support RTL layouts?

I did a quick check by setting `<html dir="RTL">` in the chrome developer console, and it looks like the UI behaves well in RTL!

PR #293, opened 2022-10-20
Add dataset metadata files to package

smusoke:
We will want to pull these dataset metadata files from our translation pipeline in the `code-dot-org` repo. So we are including them in the package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@code-dot-org/ml-playground",
-  "version": "0.0.39",
+  "version": "0.0.40",
   "main": "dist/main.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


smusoke:
> Out of curiosity, how is this different to the package from which we [load them](https://github.com/code-dot-org/ml-playground/blob/1d7063c2d799cc81212eae4db2d236f784741ff2/src/UIComponents/SelectDataset.jsx#L50) at runtime.

@breville Other than that both are referencing the same files, this doesn't change anything outside of when this repo is packaged by npm the files will be exposed.

Here is the [other PR](https://github.com/code-dot-org/code-dot-org/pull/48767/files#diff-4d25ecd2b4993515de3eddddcfc627a6f492ae0cfade490aa3b607491f6cf680R61) in our main repo that will be to pull them into our sync.

breville (approved):

PR #294, opened 2022-10-25
Move `yarn` to devDependencies

smusoke:
As title states. Our dependency on `yarn` doesn't follow Node convention. `devDependency` is a better fit being that `yarn` falls under "a package that's only needed for local development and testing".

Here is [slack thread](https://codedotorg.slack.com/archives/C0T0PNR0D/p1666651217612399) for reference.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@code-dot-org/ml-playground",
-  "version": "0.0.40",
+  "version": "0.0.41",
   "main": "dist/main.js",
   "repository": {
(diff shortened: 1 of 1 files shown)


wilkie (approved):
I like it! Thanks for taking the time to do this. 👍

fisher-alice (approved):
Thanks for doing this, Steffan. Added this PR as a reference in this jira ticket. https://codedotorg.atlassian.net/browse/SL-323

PR #295, opened 2022-10-27
Upgrading to babel 7

daynew:
There is another PR where I want to upgrade the `messageformat` package but I was blocked because it needed a newer version of babel. This change upgrades `ml-playground` to use babel 7.

This puts the babel version to be in sync with the `code-dot-org` repo.

## How I did this:
1. Run `npx babel-upgrade --write` to automatically migrate `.babelrc` and `package.json` packages to be compatible with babel 7.
2. Run `yarn install` to update the `yarn.lock` file.
3. Run `yarn start` to run the webserver and fix any report build issues or errors in the javascript console.

## Testing done
* `yarn start` and did a full run of the student experience on localhost:8080
* unit tests still pass
* Using `yarn link`, verified http://localhost-studio.code.org:3000/s/csd7-2021/lessons/9/levels/1 works.


babel.config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "presets": [
+    "@babel/preset-env",
+    "@babel/preset-react"
+  ],
+  "plugins": [
+    "add-module-exports",
+    "syntax-async-functions",
+    "@babel/plugin-syntax-dynamic-import",
```
madelynkasula, line 9: it looks like the migration tool may have done this for us, but out of curiosity, do you know why we needed to add the dynamic import plugin?

madelynkasula (approved):
left a non-blocking question, but this looks good! thank you for documenting the test steps you took!

babel.config.json
daynew: Oh, actually I don't know if we do, I was just making it the same as code-dot-org repo while I was debugging stuff. I will verify if we need this before I commit it.

PR #297, opened 2022-11-10
Add I18n for dataset metadata.

daynew:
In AI Lab, we have datasets students can use. These datasets have metadata file with content which is display to the user and needs to be translated. For example: https://github.com/code-dot-org/ml-playground/blob/main/public/datasets/billionaires.json
This metadata is sprinkled all over the app, that's why this PR is so large.

This PR also adds `default` and `scope` to the options in `I18n.t()`.
* `default`: allows a fallback string to be used if a translation isn't found.
* `scope`: allows translation keys to be nested. This makes it easier to translate nested structures like the dataset metadata JSON files.

## Links
* [JIRA](https://codedotorg.atlassian.net/browse/FND-2096)

## Testing
* `yarn test`
* manually tested on localhost.

## Video
Notes that "!!!" was prepended to all the new features being loaded by the `I18n.t()` method

https://user-images.githubusercontent.com/1372238/201016062-977c2625-c924-4484-92ec-77fb769b8650.mov

## Future work
* A follow PR will add i18n support for the dataset display names. This is being split up into a separate PR because it is outside the dataset metadata file.

### DIFF
--- a/src/UIComponents/Statement.jsx
+++ b/src/UIComponents/Statement.jsx
@@ -7,6 +7,7 @@ import { styles } from "../constants";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faTimesCircle } from "@fortawesome/free-solid-svg-icons";
 import I18n from "../i18n";
+import { getLocalizedColumnName } from "../helpers/columnDetails.js";
 
 class Statement extends Component {
   static propTypes = {
@@ -17,7 +18,8 @@ class Statement extends Component {
     labelColumn: PropTypes.string,
     selectedFeatures: PropTypes.array,
     setLabelColumn: PropTypes.func,
-    removeSelectedFeature: PropTypes.func
+    removeSelectedFeature: PropTypes.func,
+    datasetId: PropTypes.string
   };
 
   removeLabel = () => {
@@ -29,9 +31,10 @@ class Statement extends Component {
   };
 
   labelHTML = (label , currentPanel) => {
+    const localizedLabel = getLocalizedColumnName(this.props.datasetId, label);
     return (
       <div style={styles.statementLabel}>
-        {label || "____"}
+        {(label && localizedLabel) || "____"}
         {label && currentPanel === "dataDisplayLabel" && (
           <div
             id="uitest-remove-statement-label"
@@ -55,10 +58,11 @@ class Statement extends Component {
     return (
       <span>
         {selectedFeatures.map((selectedFeature, index) => {
+          const localizedName = getLocalizedColumnName(this.props.datasetId, selectedFeature);
           return (
             <span key={index}>
               <div style={styles.statementFeature}>
-                {selectedFeature}
+                {localizedName}
                 {currentPanel === "dataDisplayFeatures" && (
                   <div
                     id="uitest-remove-statement-feature"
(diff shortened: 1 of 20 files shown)


src/helpers/columnDetails.js
```diff
@@ -67,7 +69,7 @@ export function getColumnDescription(columnId, metadata, trainedModelDetails) {
     const field = metadata.fields.find(field => {
       return field.id === columnId;
     });
-    return field.description;
+    return getLocalizedColumnDescription(metadata.name, columnId, field.description)
```
wilkie, line 72: nit: missing semicolon

src/i18n.js
```diff
@@ -1,19 +1,40 @@
...
   }
-  return messages[key](options);
+  // The default value to return if no string is found for the given key
+  const defaultValue = options["default"];
+
+  const scope = options["scope"] || [];
+  let scopedMessages = messages;
+  scope.forEach(s => {
+    if (scopedMessages !== undefined) {
+      scopedMessages = scopedMessages[s]
```
wilkie, line 23: nit: missing semicolon

src/helpers/datasetDetails.js
```diff
@@ -6,17 +8,67 @@ export function getDatasetDetails(state) {
...
+      {
+        scope: ["datasets", datasetId, "card", "context"],
+        default: fallback
+      }
+    );
+  }
+  return undefined;
+}
+
+function getPotentialMisuses(state) {
```
wilkie, line 36: There is some potential DRY-ness available with a small change to merge these two functions into one to work for both `potentialMisuses` and `potentialUses`, if you'd like.

src/UIComponents/Statement.jsx
```diff
@@ -29,9 +31,10 @@ class Statement extends Component {
   };
 
   labelHTML = (label , currentPanel) => {
+    const localizedLabel = getLocalizedColumnName(this.props.datasetId, label);
     return (
       <div style={styles.statementLabel}>
-        {label || "____"}
+        {(label && localizedLabel) || "____"}
```
wilkie, line 37: Slightly confused by the order here. We would want to either render `localizedLabel` if it exists over rendering the given `label` and want `localizedLabel && label` ... but at any rate, localizedLabel will always either have a localized string OR fallback to `label` anyway?

src/selectors/visualizationSelectors.js
```diff
@@ -1,11 +1,11 @@
 import { createSelector } from 'reselect';
-import { getLabelColumn, getData, getColumnsByDataType } from "../selectors.js";
+import {getLabelColumn, getData, getColumnsByDataType, getDatasetId} from "../selectors.js";
```
wilkie, line 2: nit: inconsistent lack of space after `{` and before `}`

wilkie (commented):
Looks good with just some style/convention nits.

This work updates several places in the UI code to localize the metadata for datasets. For this, there has to be a static identifier that can be used to identify the localized strings... the "name" was used for this, but this might be localized so the `datasetId` is passed through instead. Some helpers are used with that id to localize the column names (whereas before the `id` (name) was used from the dataset data directly). That column data, instead of coming from the metadata block, comes from `getDatasetDetails`, a helper extended to get localized strings for the details block (adding `potentialUses` / `potentialMisuses` as well).

The I18n mechanism has been augmented with the added default/scope options which match their behavior in the ruby/rails i18n library. Good stuff.

The tests are thorough and test the localization paths for different datasets. All good.

src/UIComponents/Statement.jsx
daynew: yeah you are right we can just use localizedLabel instead of label.

PR #299, opened 2023-04-03
Show filter error

breville:
Now, when we receive a filter error from the server, we show a numerical code so that we can determine the type of error:

Screenshot 2023-04-03 at 12 23 32 PM

We also log error details to the console:

Screenshot 2023-04-03 at 12 24 22 PM

This change depends on https://github.com/code-dot-org/code-dot-org/pull/51084 on the server.

For now, we are exposing just enough information that we can help out via support ticket.  Going forward, we might well decide to expose more information, and indeed to give actionable feedback so that users can adjust the data on their end.

There is also the broader question of whether we should do some of this filtering at the beginning of the user experience, at least in the cases where it will lead to this model saving.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -34,6 +34,7 @@ class PanelButtons extends Component {
     startSaveTrainedModel: PropTypes.func,
     dataToSave: PropTypes.object,
     saveStatus: PropTypes.string,
+    saveData: PropTypes.object,
     isSaveComplete: PropTypes.func,
     shouldDisplaySaveStatus: PropTypes.func
   };
@@ -52,16 +53,38 @@ class PanelButtons extends Component {
     }
   };
 
-  localizedSaveMessage = (saveStatus) => {
-    return I18n.t(`saveStatus_${saveStatus}`)
+  localizedSaveMessage = saveStatus => {
+    return I18n.t(`saveStatus_${saveStatus}`);
+  };
+
+  saveDataMessage = saveData => {
+    // The list of known error types from share_filtering.rb.
+    const errorTypes = ["email", "address", "phone", "profanity"];
+
+    if (!saveData) {
+      return "";
+    }
+
+    const index = errorTypes.indexOf(saveData.type);
+    if (index !== -1) {
+      return ` (${index})`;
+    } else {
+      return "";
+    }
   };
 
   render() {
-    const { panelButtons, saveStatus } = this.props;
+    const { panelButtons, saveStatus, saveData } = this.props;
 
-    let loadSaveStatus= this.props.isSaveComplete(saveStatus)
-      ? this.localizedSaveMessage(saveStatus)
-      : ( <FontAwesomeIcon icon={faSpinner} />);
+    let loadSaveStatus = this.props.isSaveComplete(saveStatus) ? (
+      this.localizedSaveMessage(saveStatus)
+    ) : (
+      <FontAwesomeIcon icon={faSpinner} />
+    );
+
+    let loadSaveData = this.props.isSaveComplete(saveStatus)
+      ? this.saveDataMessage(saveData)
+      : undefined;
 
     return (
       <div style={styles.navigationButtonsContainer}>
@@ -85,7 +108,10 @@ class PanelButtons extends Component {
 
         {this.props.shouldDisplaySaveStatus(saveStatus) &&
           this.props.currentPanel === "saveModel" && (
-            <span style={styles.modelSaveMessage}>{loadSaveStatus}</span>
+            <span style={styles.modelSaveMessage}>
+              {loadSaveStatus}
+              <span style={styles.modelSaveDataMessage}>{loadSaveData}</span>
+            </span>
           )}
 
         {panelButtons.next && (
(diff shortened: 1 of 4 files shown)


breville:
> Looks good to me! A question we can take to Slack - what's the list of error codes and what they match to? I'd want to add that somewhere outside of the code so curriculum / customer support can access it.

Good question.  It's just an ordinal index for the list [here](https://github.com/code-dot-org/ml-playground/pull/299/files#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67R62), specifically:

`0`: email
`1`: address
`2`: phone
`3`: profanity

> Also, flagging that since AI Lab is now translatable, I think that means we should give #platform-team a heads up on the change to this string? Apologies if you already did that.

Absolutely.  I tagged @daynew for this reason, and hopefully he can tag anybody else who should also take a look.

molly-moen (approved):

src/constants.js
```diff
@@ -952,6 +953,12 @@ export const styles = {
     width: 500
   },
 
+  modelSaveResponseDataMessage: {
+    display: "inline-block",
+    opacity: 0.5,
+    padding: "0 5px"
+  },
+
```
jmkulwik, line 961: Maybe a naive question: Why are css styles in a constants.js file?

src/App.js
```diff
@@ -52,16 +53,38 @@ class PanelButtons extends Component {
...
+    // The list of known error types from share_filtering.rb.
+    const errorTypes = ["email", "address", "phone", "profanity"];
+
+    if (!saveResponseData) {
+      return "";
+    }
+
+    const index = errorTypes.indexOf(saveResponseData.type);
+    if (index !== -1) {
+      return `(${index})`;
```
jmkulwik, line 70: Why use an error code rather than telling them the type of error?

src/constants.js
breville: I don't think any particular reason, except apparently some early ones got added there and then it just stuck.  We could fix, though I'd prefer to isolate that to its own change.

src/redux.js
```diff
@@ -602,7 +605,8 @@ export default function rootReducer(state = initialState, action) {
   if (action.type === SET_SAVE_STATUS) {
     return {
       ...state,
-      saveStatus: action.status
+      saveStatus: action.status,
+      saveResponseData: action.data
```
jmkulwik, line 609: Probably out of scope: Would it be worth breaking this into multiple redux files? This file's getting pretty big.

breville: It could be interesting, but definitely out of scope for this change.

src/App.js
breville: My idea was that we would start with an incremental change that would let us continue to handle these via support ticket, but we can ask for this number to begin diagnosis.  It seems relatively rare, and will help us understand the nature of problems occurring in the wild.  We can also keep an eye on the new Firehose events to get the relevant data on our end, support incident or not.  

As time goes on, we can certainly add more actionable feedback, but we'll need to figure out the right user experience for that.

mikeharv (approved):
Thanks for coming up with the solution here! It looks good. I left some comments, mostly questions for my own knowledge and one about a potentially repeated condition.

src/App.js
```diff
@@ -52,16 +53,38 @@ class PanelButtons extends Component {
...
 
-  localizedSaveMessage = (saveStatus) => {
-    return I18n.t(`saveStatus_${saveStatus}`)
+  localizedSaveMessage = saveStatus => {
+    return I18n.t(`saveStatus_${saveStatus}`);
+  };
+
+  saveResponseDataMessage = saveResponseData => {
+    // The list of known error types from share_filtering.rb.
+    const errorTypes = ["email", "address", "phone", "profanity"];
```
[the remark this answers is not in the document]
breville: It would be handy to have a single source of truth for these, and we do solve that in the main repo by having some constants in JS that are generated from Ruby.  However, I'm not sure of a great automated way to get this from the main repo across to this standalone one.

src/App.js
```diff
@@ -52,16 +53,38 @@ class PanelButtons extends Component {
...
+    let loadSaveStatus = this.props.isSaveComplete(saveStatus) ? (
+      this.localizedSaveMessage(saveStatus)
+    ) : (
+      <FontAwesomeIcon icon={faSpinner} />
+    );
 
-    let loadSaveStatus= this.props.isSaveComplete(saveStatus)
-      ? this.localizedSaveMessage(saveStatus)
-      : ( <FontAwesomeIcon icon={faSpinner} />);
+    let loadSaveResponseData = this.props.isSaveComplete(saveStatus)
```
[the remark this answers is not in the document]
breville: It's a fair point.  One downside is that if we are using `const` (which I think we should) then I think we lose the ability to do that if we have to declare the variable outside of the conditional body that possibly sets it.

src/App.js
```diff
@@ -85,7 +108,12 @@ class PanelButtons extends Component {
 
         {this.props.shouldDisplaySaveStatus(saveStatus) &&
           this.props.currentPanel === "saveModel" && (
-            <span style={styles.modelSaveMessage}>{loadSaveStatus}</span>
+            <div style={styles.modelSaveMessage}>
+              {loadSaveStatus}
+              <div style={styles.modelSaveResponseDataMessage}>
+                {loadSaveResponseData}
```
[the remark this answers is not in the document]
breville: With the current code, it will actually `""` if there's nothing to show, but I can adjust so that we don't create the `div` at all in that case.

PR #300, opened 2023-04-05
Add a prettier config

breville:
This just adds a `prettier` config to this repo, with settings that I found maintained the status quo, which is handy for making small changes.


molly-moen (approved):
🎉

PR #301, opened 2023-04-19
Update toy datasets with missing name property

bencodeorg:
We noticed that descriptions were not showing up for a few datasets -- traced to a missing `name` property.

## Links

- Slack discussion: [link](https://codedotorg.slack.com/archives/C03DBDN67B7/p1681910571638339)

## Testing story

Confirmed by loading these dataset descriptions at `/s/csd7-2022/lessons/19/levels/1` locally in `code-dot-org`, and saw them displaying correctly. Also confirmed they did not load correctly prior to this change.

image

### DIFF
--- a/public/datasets/club_biased_toy.json
+++ b/public/datasets/club_biased_toy.json
@@ -1,4 +1,5 @@
 {
+  "name": "club_biased_toy",
   "card": {
     "name": "Nico's Club Survey",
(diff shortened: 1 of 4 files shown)


dancodedotorg (approved):

molly-moen (approved):
I don't know much about this but this looks reasonable to me!

PR #307, opened 2026-04-09
Switch to React functional components

wilkie:
Fairly straight-forward conversion. Though it chose to make them in the form `function MyComponent(...) {}` and I would prefer `const MyComponent = ({...}: MyComponentProps) => {...}` instead. I will convert this in the TypeScript path.


package.json
```diff
@@ -32,26 +32,27 @@
...
     "identity-obj-proxy": "^3.0.0",
     "jest": "29",
     "messageformat": "2.3.0",
     "ml-knn": "^3.0.0",
     "query-string": "4.1.0",
-    "react": "~15.4.0",
-    "react-dom": "~15.4.0",
+    "react": "18",
+    "react-dom": "18",
     "react-papaparse": "^3.8.0",
```
stephenliang: These are exact pins, which we shouldn't use. Although not necessarily important because when we bring it to `frontend`, we'd use the catalog version instead. I'd prefer we use a more relaxed semver, but I leave that up to you if we are going to inline to the main repo soon.

src/UIComponents/TrainModel.jsx
```diff
@@ -16,168 +16,168 @@ import I18n from "../i18n";
...
-  componentWillUnmount = () => {
-    if (this.state.animationTimer) {
-      clearInterval(this.state.animationTimer);
-      this.setState({ animationTimer: undefined });
-    }
-  };
+    return () => {
+      clearInterval(animationTimer);
+    };
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
stephenliang: nit: Comment why or refactor to avoid?

stephenliang (approved):

src/UIComponents/TrainModel.jsx
wilkie: Wrapped it in a ref to avoid needing it as a dependency.

package.json
wilkie: Loosened the versions. Not a problem when we port it over, but good to do.

PR #308, opened 2026-04-09
Convert to TypeScript

wilkie:
Uses the strict TypeScript config in `/frontend/packages/lint-config` to get this repo locked to the same TypeScript and TS linting situation as the main repo.

This also converts `function MyComponent({}) { }` into `const MyComponent = ({}: MyComponentProps) => { }`

Snuck an upgrade to fontawesome which just updated the package.json/yarn.lock.

https://github.com/user-attachments/assets/b4463321-8615-4ab9-9545-0f1571928a57


src/UIComponents/DataCard.tsx
```diff
@@ -29,12 +37,12 @@ function DataCard({ name, metadata, datasetDetails, dataLength, removedRowsCount
         <ScrollableContent>
           {card && (
             <div>
-              <div style={styles.cardRow}>{datasetDetails.description}</div>
+              <div style={styles.cardRow}>{datasetDetails!.description}</div>
```
stephenliang: Technically equivalent, but leads me to wonder if there's a code smell in which we can early handle falsy `datasetDetails` or make it required rather than an optional field. Same applies to the other fields.

src/UIComponents/DataTable.tsx
```diff
@@ -153,41 +172,24 @@ function DataTable({
...
     currentColumn: state.currentColumn,
     highlightColumn: state.highlightColumn,
     currentPanel: state.currentPanel
   }),
-  dispatch => ({
-    setCurrentColumn(column) {
-      dispatch(setCurrentColumn(column));
+  (dispatch: Dispatch) => ({
+    setCurrentColumn(column: string | undefined) {
+      dispatch(setCurrentColumn(column as string));
```
stephenliang, line 196: nit: this indicates that the downstream user should also be `string | undefined` to avoid the cast.

src/UIComponents/GenerateResults.tsx
```diff
@@ -49,10 +67,10 @@ function GenerateResults({ data, readyToTrain, labelColumn, selectedFeatures, in
     return () => {
       clearInterval(animationTimer);
     };
-  }, [data.length]);
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
stephenliang, line 76: nit: prefer to know why or preferably avoid

src/UIComponents/ModelCard.tsx
```diff
@@ -90,30 +98,30 @@ function ModelCard({ trainedModelDetails, selectedFeatures, percentCorrect, labe
           <h5 style={styles.modelCardHeading}>{I18n.t("modelCardLabel")}</h5>
           <div style={styles.modelCardContent}>
             <p style={styles.bold}>{localizedLabel}</p>
-            {label.description && <p>{label.description}</p>}
-            {!label.values && (
+            {label!.description && <p>{label!.description}</p>}
```
stephenliang: optional: Same comment that this is technically equivalent but leads me to wonder if we should early exit or handle it some other way for a better UX.

stephenliang (approved):
Very minor nits, all of which are probably considered stylistic or out of scope for this PR (error handling scenarios). However I do think we can possibly avoid a type cast in the dispatcher.

stephenliang (commented):
Note: I think you need to also add a `typecheck` script. If I run typecheck manually, it seems to error.

```
npx tsc
```

throws quite a number of type errors.

src/UIComponents/DataCard.tsx
wilkie: Yeah, good to call out. I dislike seeing any `!` and I'll look at smoothing the types out.

src/UIComponents/GenerateResults.tsx
wilkie: Already taken care of in the other PR. Smoothed it out with a ref.

wilkie:
I have rebased, added a `yarn typecheck` command, added the necessary type declarations, and addressed many of the weird situations from the old codebase where functions return `null` only to be assumed not-null by consumers later that the TypeScript port revealed (and the AI just... allowed for some reason.) I'm testing it and will push it up and then probably merge it and move toward the localization fixes we need for CSD and also introducing this to the main repo.

wilkie:
Added prettier config to match the main repo and just had it 'fix' everything. Now it should conform 99% to what the frontend packages directory wants.

wilkie:
I confirmed that all of the modes work, the metrics fire, and the animations seem to be correct. No major issues, certainly.

PR #309, opened 2026-04-09
Localize data rows

wilkie:
Adds a helper to handle localizing the data values in the datasets.

This looks for a localization string in the `datasets.{datasetId}.values` object where the English string is the key. So, 'Yes' would be: `"datasets": {"abalone": {"values": "Yes": "Yes"}}}`, etc.


stephenliang (approved):
I'd prefer we also add unit tests for this

PR #315, opened 2026-05-26
Fixes issues with redux.

wilkie:
We need to make this package's redux the same version as the apps main.

We also need to make sure we don't update state inside a reducer and instead queue a microtask. Otherwise the app crashes when selecting datasets.

This problem was discovered on the last level of ailab in CSD Unit 7. It's a rare interaction. This will fix the last remaining issue that I can see at the moment such that we can enable this build on production.


src/redux.ts
```diff
@@ -416,10 +416,20 @@ export default function rootReducer(
...
-        null,
-      );
+      // Reducer must stay pure: the consumer-supplied callback dispatches
+      // into its own redux store, which would interleave React commits
+      // (and a getState cascade) into this dispatch and trip the
+      // "getState() while reducer is executing" guard. Defer to a
+      // microtask so the dispatch fully unwinds before the callback fires.
+      if (state.instructionsKeyCallback) {
+        const callback = state.instructionsKeyCallback;
+        queueMicrotask(() =>
```
sanchitmalhotra126, line 426: Nice find! Ideally we wouldn't have a callback in here at all and/or move this to a thunk, but that can happen later

sanchitmalhotra126 (approved):

src/redux.ts
wilkie: Yep. I don't like a bit of how this is working from a certain standpoint, but the key word there is it is _working._ When we port it into the actual main repo, we have much more opportunity to clean it up. I'm not sure if ml-playground itself really needs its own redux to do what it is doing anyway.

